PR #1, opened 2020-04-04
Add "blank" Playground app from template

ryantrem:
This is just a "blank" React Native (TypeScript) project created with:
npx react-native init Playground --template react-native-template-typescript


ryantrem:
This is the first of three initial commits to create a reasonable git history from the work that has been done so far. The three commits will be:
1. A "blank" Playground React Native app
2. A "blank" react-native-babylon React Native module
3. A fleshed out initial implementation of react-native-babylon module and Playground app to test it

PR #3, opened 2020-04-07
Adding initial implementation for Android integration

ryantrem:
This adds the initial integration of Babylon Native into React Native for Android specifically. The changes are described in [readme.md](https://github.com/ryantrem/BabylonReactNative/blob/782d833db911d3372b6a5e5b6d53ea6e6159386a/README.md).

This is the third of three initial commits to create a reasonable git history from the work that has been done so far. The three commits will be:

1. ~~A "blank" Playground React Native app~~ ✔
1. ~~A "blank" react-native-babylon React Native module~~ ✔
1. A fleshed out initial (Android) implementation of react-native-babylon module and Playground app to test it


Apps/Playground/App.tsx
```diff
@@ -1,118 +1,102 @@
...
+    };
 
-declare var global: {HermesInternal: null | {}};
+    const timerHandle = setTimeout(() => {
+      setFps(engine.getFps());
+    }, 1000);
+
+    return () => {
+      clearTimeout(timerHandle);
+    };
```
bghgary, line 49: Are you trying to show the FPS regularly? I'm not sure I understand what this is doing.

README.md
```diff
@@ -0,0 +1,140 @@
+# Babylon React Native
```
bghgary, line 1: We probably need to follow Microsoft open source guidelines here as the plan is to do this similar to Babylon Native repo.

Apps/Playground/android/build.gradle
```diff
@@ -3,16 +3,17 @@
 buildscript {
     ext {
         buildToolsVersion = "28.0.3"
-        minSdkVersion = 16
+        minSdkVersion = 24
```
syntheticmagus, line 6: Is this required? I guess maybe it's for ARCore, but changing the minimum from 16 to 24 means about 30% of Android devices in the wild won't be able to use this at all (if it works the way I think it works).

https://developer.android.com/about/dashboards/

Apps/Playground/node_modules/react-native-babylon/android/src/main/cpp/BabylonNativeInterop.cpp
```diff
@@ -0,0 +1,104 @@
...
+        {
+            auto width = static_cast<size_t>(ANativeWindow_getWidth(windowPtr));
+            auto height = static_cast<size_t>(ANativeWindow_getHeight(windowPtr));
+            ReinitializeNativeEngine(*m_env, windowPtr, width, height);
+        }
+
+    private:
+        using looper_scheduler_t = arcana::looper_scheduler<sizeof(std::weak_ptr<Napi::Env>) + sizeof(std::function<void(Napi::Env)>)>;
+        std::shared_ptr<looper_scheduler_t> m_looper_scheduler;
+        std::shared_ptr<Napi::Env> m_env;
```
syntheticmagus, line 81: I'm not sure either of these should be shared pointers. The `Env` almost certainly doesn't need to be; it can just be captured by copy in the `JsRuntime`'s dispatch function, which will then hold onto it forever. I'm less sure about the looper. Does it need to be owned by this? Or can it just be `std::move`ed into the dispatch function so that the `JsRuntime` can own it instead of just referencing it?

Apps/Playground/App.tsx
ryantrem: Yes, it was a last minute addition that I messed up. Updated to `setInterval`.

README.md
ryantrem: I'm not sure what that means. Is there something you want me to change now?

bghgary: No, not yet.

Apps/Playground/android/build.gradle
ryantrem: I ran into something that required me to bump it to 21, and then something else that went to 24. I think it was the use of CompletionStage<T> or something, but I can't remember for sure. I know I did this for a reason though, so my vote would be to wait until we have evidence that consumers of the lib want it, and then we can re-evaluate.

Apps/Playground/node_modules/react-native-babylon/android/src/main/cpp/BabylonNativeInterop.cpp
ryantrem: Good point. I need to leave the looper_scheduler as a shared_ptr because looper_scheduler is not copyable, but the shared_ptr can be owned by the dispatch function.

PR #9, opened 2020-05-15
Integrate XR support

ryantrem:
This change brings XR support into the react native layer. One notable change here is with the camera permissions request code I added in the JS layer. See the comments for some explanation, and let me know if anyone has thoughts on the way I currently just "inject" the permissions check into the existing XR session initialization function.

NOTE: This PR depends on https://github.com/BabylonJS/BabylonNative/pull/264. Once this PR is completed, I will update the Babylon Native submodule to master.

### DIFF
--- a/Apps/Playground/node_modules/react-native-babylon/android/src/main/cpp/BabylonNativeInterop.cpp
+++ b/Apps/Playground/node_modules/react-native-babylon/android/src/main/cpp/BabylonNativeInterop.cpp
@@ -36,7 +36,7 @@ namespace Babylon
     public:
         // This class must be constructed from the JavaScript thread
         Native(JSGlobalContextRef jsContext, ANativeWindow* windowPtr)
-        : m_env{ Napi::Attach<JSContextRef>(jsContext) }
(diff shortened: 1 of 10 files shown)


Apps/Playground/node_modules/react-native-babylon/android/src/main/cpp/BabylonNativeInterop.cpp
```diff
@@ -64,6 +68,8 @@ namespace Babylon
         ~Native()
         {
             Plugins::NativeEngine::DeinitializeGraphics();
+            // TODO: Figure out why this causes the app to crash
+            //Napi::Detach(m_env);
```
bghgary, line 72: Does this still happen?

ryantrem: Yes, this is actually why when I tried to integrate your JSC changes it was crashing in some bgfx cleanup ('cause I also had this line). I just don't want to dig into it now given the JSI work.

Issue #15, opened 2020-05-28
iOS integration


ryantrem:
Remaining tasks for this:

- [x] Handle screen rotation (e.g. resize)
- [x] Handle reinstantiation of EngineView
- [x] Handle reinstantiation of Engine
- [x] Handle remote debugging (e.g. don't crash, return 0)
- [x] Handle suspend/resume
- [x] Handle runtime warning about initializing graphics on JS thread
- [x] Forward through touch/pointer input
- [x] Figure out what to do about bitcode (enabled by default for React Native apps, disabled in our cmake build for Babylon Native)
- [x] Make release (profile) build work

ryantrem:
Remaining tasks for this:

- [ ] Handle screen rotation (e.g. resize)
- [ ] Handle reinstantiation of EngineView
- [ ] Handle reinstantiation of Engine
- [ ] Handle remote debugging (e.g. don't crash, return 0)
- [ ] Handle suspend/resume
- [ ] Handle runtime warning about initializing graphics on JS thread
- [ ] Forward through touch/pointer input
- [ ] Figure out what to do about bitcode (enabled by default for React Native apps, disabled in our cmake build for Babylon Native)
- [ ] Make release (profile) build work

PR #30, opened 2020-06-17
Initial iOS integration

ryantrem:
This is the initial work to get iOS integration working. There are a bunch of known things still left, which I've listed off in issue #15, but this change brings basic functionality. The big parts of this change are:

- Minimal changes to get the existing iOS code building (there were issues before I even started adding new code).
- Add a `Babylon::Native` class that mirrors what we did on the Android side (ties together the set of underlying Babylon Native constructs we want to use in the React Native integration).
- Add a `CMakeLists.txt` that compiles `Babylon::Native` (and all its dependencies on Babylon Native).
- Add instructions to manually run `CMake` on this new `CMakeLists.txt` to generate an XCode project.
- Add the generated XCode project to the Playground's XCode workspace.
- Add a ton of static libs to be linked by the Playground XCode project.
- Add an interop layer that mirrors the interop layer on Android.
- Add a custom React Native module that mirrors the Android side.
- Added a native `EngineViewManager` that mostly mirrors what we do on the Android side. One difference here is that there doesn't seem to be a way to get onto the JS thread directly from the native code, so now early in the initialization the JS code makes a synchronous call into the native code so we can capture the `CFRunLoop` for the JS thread and later schedule work on this thread.
- Updated JavaScript `EngineView` to call this new synchronous initialization function.
- Updated to the latest BabylonNative submodule, which in turn includes the latest arcana.cpp, which includes the new `run_loop_scheduler` (based on `CFRunLoop`).
- Added the `JSCRuntime.h` that @syntheticmagus created that gives us access to the underlying JSC `JSGlobalContextRef`. In a different change I will update the Android side to deal with the JS context similarly to how I did it here for iOS, which will make for a smoother transition away from the custom `JSCRuntime.h` to just using JSI directly.


Apps/Playground/node_modules/@babylonjs/react-native/ios/EngineViewManager.mm
```diff
@@ -0,0 +1,72 @@
...
+        initialized = false;
+        
+        super.translatesAutoresizingMaskIntoConstraints = false;
+        super.colorPixelFormat = MTLPixelFormatBGRA8Unorm_sRGB;
+        super.depthStencilPixelFormat = MTLPixelFormatDepth32Float;
+    }
+    return self;
+}
+
+- (void)setBounds:(CGRect)bounds {
```
CedricGuillemet, line 31: Might use viewDidLayout instead?

ryantrem: As far as I know that only exists on a `ViewController`, which is not used in React Native (you use `UIView`s directly).

PR #31, opened 2020-06-20
Various iOS integration fixes

ryantrem:
This fixes about half the remaining issues listed in #15. The big changes here are:

- Get release build working - `libbx` and `libbgfx` have different names for debug and release (`d` is appended in debug). To fix this, these libs are directly configured as linker options that differ between debug and release, which seems to be the only way to configure this in XCode.
- Handle engine reinstantiation - this happens when the React Native app reloads the JS context. This only required adding a call to `DeinitializeGraphics` when the native instance is destroyed.
- Don't crash when remote debugging is enabled - don't try to make a synchronous JS call (it's not possible when remote debugging is enabled), change the interop layer to key off of the `RCTBridge` rather than the `jsi::Runtime`, and handle the case where the `jsi::Runtime` hanging off the `RCTBridge` is `null` (this happens when remote debugging is enabled).
- Forward touch input - this was just plumbing through touch input through the interop layer and into the Babylon native layer.

Other changes worth mentioning:

- I changed the structure of the code to handle view reinstantiation and view resizing, but both of these things cause crashes, so some new code is commented out until this is resolved.
- I added some code to call `InitializeGraphics` on the UI thread (there is a runtime warning about this currently, and the regular Babylon Native Playground runs this on the UI thread, so I think this is what we need to do), but this also causes a crash so it is commented out until this is resolved.
- I moved some of the `setView` logic out of `EngineView` and into the interop layer to be more consistent with what we do on Android and to simplify the code in `EngineView` a bit.
- I hid the BabylonNative.h file from the POD since it is already visible in the generated ReactNativeBabylon project.


Apps/Playground/node_modules/@babylonjs/react-native/ios/BabylonNative.cpp
```diff
@@ -38,6 +38,10 @@ namespace Babylon
     Native::Native(facebook::jsi::Runtime* jsiRuntime, void* windowPtr, size_t width, size_t height)
         : m_impl{ std::make_unique<Native::Impl>(jsiRuntime) }
     {
+//        dispatch_sync(dispatch_get_main_queue(), ^{
+//            Plugins::NativeEngine::InitializeGraphics(windowPtr, width, height);
```
Alex-MSFT, line 42: ?

Apps/Playground/node_modules/@babylonjs/react-native/ios/BabylonNative.cpp
```diff
@@ -59,6 +63,11 @@ namespace Babylon
         m_impl->nativeInput = &Babylon::Plugins::NativeInput::CreateForJavaScript(m_impl->env);
     }
 
+    Native::~Native()
+    {
+        Plugins::NativeEngine::DeinitializeGraphics();
```
Alex-MSFT, line 68: Does this make sense if initialize graphics is commented out

Apps/Playground/node_modules/@babylonjs/react-native/ios/BabylonNative.cpp
ryantrem: Sorry for the confusion, this is here and commented out because we are supposed to call `InitializeGraphics` from the UI thread, but for some reason it makes the app crash. I added a TODO comment (and this is still an unchecked todo in the GitHub issue for iOS React Native integration). I just need to dig into it more, or get some help from @CedricGuillemet.

PR #42, opened 2020-06-30
Add a PackageTest app

ryantrem:
The Playground app makes it possible to quickly iterate on the @babylonjs/react-native package in the context of an app, but it doesn't help validate the package itself. With this change, I'm adding a PackageTest app that consumes the actual @babylonjs/react-native package. In the checked in form, it references a locally built package (`.tgz` file), but you can easily switch it to reference the package from npmjs.org as well. This is mostly in preparation for creating binary packages, which we will need to test more rigorously than the source based packages (which are intended to be identical to the checked in package source).

I also fixed a few package creation issues and documentation issues I found while trying to actually use the package in the PackageTest app.

NOTE: The first commit is just the default React Native app generated the the React Native CLI, so the best way to review this PR is to just look at the last three commits (that is, all commits but the first). That will show you the diff between a default blank RN app and the actual PackageTest app I setup.


Apps/PackageTest/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "scripts": {
+    "android": "react-native run-android",
+    "ios": "react-native run-ios",
+    "start": "react-native start",
+    "test": "jest",
+    "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
+  },
+  "dependencies": {
+    "@babylonjs/core": "^4.2.0-alpha.21",
+    "@babylonjs/react-native": "file:../Playground/node_modules/@babylonjs/react-native/babylonjs-react-native-0.0.1.tgz",
```
ryantrem, line 14: This is the important part of this test app. All you need to do is `npm pack` in the `Apps/Playground/node_modules/@babylonjs/react-native` directory, then follow the normal steps in "initializing" the app (e.g. `npm install`, `pod install`, etc.).

PR #45, opened 2020-07-08
PR builds and binary package builds for releases

ryantrem:
The main change here is to update `publish.yml` to now prep the environment for building source code, and then actually building the source using `gulp`, and finally publishing the binary package (instead of the source package). I've already created a release to test the full flow and the created package (https://www.npmjs.com/package/@babylonjs/react-native). I have one more package being built and published now, which updates the README.md to remove the steps that are no longer needed since the source is pre-compiled.

- Update publish.yml to install Ninja, do a Yarn install on the Playground app (since we build it to produce the libBabylonNative.so files for Android), do a NPM install in the Package directory (to install a few tools and libraries needed for the Gulp build), then actually run the Gulp build, and finally publish the pre-compiled package instead of the source package.
- Update the build.gradle files to switch to a different version of the NDK which is pre-installed on GitHub build agents (and update the associated documentation).
- Update the package README.md to remove the steps that are no longer necessary with the pre-compiled package.

Additionally, since building the code is now relatively trivial, I am also adding `pr.yml` to facilitate gated PR builds. This duplicates steps from `publish.yml` as there doesn't seem to be any way to share a set of steps across jobs with GitHub Actions/Workflows, but the steps are really simple so meh.


Apps/Playground/node_modules/@babylonjs/react-native/android/build.gradle
```diff
@@ -14,7 +14,7 @@ def DEFAULT_COMPILE_SDK_VERSION = 28
 def DEFAULT_BUILD_TOOLS_VERSION = '28.0.3'
 def DEFAULT_MIN_SDK_VERSION = 18
 def DEFAULT_TARGET_SDK_VERSION = 28
-def DEFAULT_NDK_VERSION = '21.0.6113669'
+def DEFAULT_NDK_VERSION = '21.3.6528147'
```
Alex-MSFT, line 17: Just curious, any reason we needed to update the NDK version?

Alex-MSFT: Ah nevermind I reread the description, this is what is preinstalled on the Github build machines.  This version seems relatively new, are we going to need to regularly update the NDK version?

ryantrem: From the PR description :) *"Update the build.gradle files to switch to a different version of the NDK which is pre-installed on GitHub build agents"*

Alex-MSFT: Yeah I saw that, I was also wondering if this version is a moving target, are we going to need to periodically keep track and update this to keep our PR builds working?

ryantrem: I don't know. I tried not specifying it, but then it just picked some default that is not installed on my machine or on the build agents. :(
As far as I can tell, there is not a way to say "use latest" or anything like that.

PR #47, opened 2020-07-10
Use NAPI over JSI over JSC

bghgary:
This change switches Babylon Native to use NAPI over JSI.

Notes:
* There are outstanding issues with the JSI implementation over JSC and thus we currently have a custom JSCRuntime.h and JSCRuntime.cpp to compensate. These changes will need to be upstreamed to Facebook in the future.
* Not all of the NAPI contract is implemented yet, but enough is implemented so that the default app runs with XR support.
* Returning NAPI types that don't match exactly with the return type (e.g. returning a Napi::Object for a function that returns Napi::Value will result in a C++ warning only when using NAPI/JSI). You must add an std::move in this situation as will be indicated by a note following the warning.
* Performance is slower with JSI due to various reasons. This will be documented in the future along with measurements.
* Code has been tested with Hermes which adds additional slowness.

TODO
- [x] Test on iOS (@ryantrem Can you help with this?)
- [x] Merge Babylon Native counterpart


Apps/Playground/node_modules/@babylonjs/react-native/JSCRuntime.cpp
```diff
@@ -0,0 +1,1448 @@
+// NOTE: This file was copied from the JSCRuntime.cpp included in React Native 0.62.1 with modifications in order to:
+//   1. Access the underlying JSC JSGlobalContextRef.
+//   2. Create a custom JSCRuntime with bug fixes and performance improvements.
```
bghgary, line 3: Note that latest version of JSCRuntime has some of the perf issues already done.

PR #48, opened 2020-07-15
Upgrade React Native (Playground & PackageTest) to 0.63.0

ryantrem:
This change upgrades the Playground and PackageTest apps to React Native 0.63.0. I decided to hold off on updating the actual @babylonjs/react-native dependencies on React Native for now since we are currently relying changes local to this repo that have not made it into a React Native release yet (e.g. JSCRuntime.cpp). Most of these changes are "boiler plate," in that I just recreated the apps from scratch and pulled forward the few customizations we'd previously made. Some notable parts of this change:
- I updated the root .gitignore as some additional cleanup in this set of changes (I believe all the additions I made to .gitignore are generated files).
- I noticed that bitcode was still not being enabled on our libs (I had tested this before, so I'm not sure why it's not working now). I put a comment in the CMakeLists.txt and for now updated the readme and the build script to enable bitcode in the cmake command line.
- I updated the gulp build to validate the set of files that are being published for the package. This is to help prevent us from accidentally publishing a broken package.


Apps/PackageTest/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.0.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.2-all.zip
```
Alex-MSFT, line 3: If you upgrade this to 6.3 we can use JDK 14

Apps/Playground/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Mon Mar 23 16:54:28 PDT 2020
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.2-all.zip
```
Alex-MSFT, line 3: Same here

Apps/PackageTest/android/gradle/wrapper/gradle-wrapper.properties
ryantrem: Updated, hopefully the lab build succeeds. 🤞🏻

PR #58, opened 2020-07-28
Update Babylon Native and Babylon.js to latest

ryantrem:
- Updating Babylon Native to get the BGFX fix for re-instantiating the EngineView on iOS (as described in #15) (and verified this is fixed).
- Updating Babylon.js to get the 32-bit index buffer fix (as described in https://github.com/babylonjs/babylonnative/issues/321) (did not verify this fix, but I think @RonnieAshlock did).

### DIFF
--- a/Apps/Playground/package-lock.json
+++ b/Apps/Playground/package-lock.json
@@ -864,9 +864,9 @@
       }
     },
     "@babylonjs/core": {
-      "version": "4.2.0-alpha.21",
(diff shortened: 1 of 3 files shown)


RonnieAshlock:
The offending test mesh:

https://github.com/RonnieAshlock/TestAssets/blob/master/GLB/WarehouseShelf.glb

Loaded as expected and is no longer malformed. Looks correct to me!

Issue #60, opened 2020-07-30
SceneLoader.ImportMeshAsync "hangs" until the display is tapped


ryantrem:
When calling `SceneLoader.ImportMeshAsync`, it "hangs" (does not complete) until the screen is tapped. You can tap anywhere, even on a different UI element (like a React Native button) and suddenly the function completes and the app runs. This repros on both Android and iOS.

ryantrem:
I did some more testing and it seems like we can get stuck on one of several (any?) calls into Babylon Native that end up in native code (until you tap the display, then things get unstuck). The behavior is very similar on both Android and iOS. Need more digging and debugging.

ryantrem:
Getting a bit more clarity on this... it seems like with "slow" async functions, the looper goes to sleep, and is woken up to call resolve, but resolve does not synchronously run the js continuations (`then`s). Somehow those get queued, but the looper is not worken up so they just sit in the queue until something wakes up the looper (e.g. user input, looper based timers, etc.). I need to understand better how promises work under the hood, particularly in the context of a JS engine like JavaScriptCore that does not manage its own threads or work queues but rather expects the caller to do this.

PR #63, opened 2020-08-01
Workaround for promise continuations not executing

ryantrem:
The main change here is a workaround for promise continuations not executing. This happens because the looper/runloop goes to sleep and nothing wakes it up. Adding a repeating timer wakes it up. This is not a good solution at all, but a passable workaround to unblock usage. If we have a lot of fast async functions using promises, this will be a bottleneck in performance, so we should definitely dig into #60 more.

I also updated the README.md to include instructions on switching from the binary npm package to installing a symbolic link to a local clone of the BabylonReactNative repo to make it easy for people to test/debug in the context of their app (I needed to do this, and it's easy, but definitely worth documenting).

### DIFF
--- a/Modules/@babylonjs/react-native/EngineHook.ts
+++ b/Modules/@babylonjs/react-native/EngineHook.ts
@@ -66,7 +66,48 @@ export function useEngine(): Engine | undefined {
                     }
                 }
 
-                setEngine(engine = new NativeEngine());
+                engine = new NativeEngine();
+
+                // NOTE: This is a workaround for https://github.com/BabylonJS/BabylonReactNative/issues/60
+                let heartbeat: NodeJS.Timeout | null;
+                const startHeartbeat = () => {
+                    if (!heartbeat) {
+                        heartbeat = setInterval(() => {}, 10);
+                    }
+                };
+                const stopHeartbeat = () => {
+                    if (heartbeat) {
+                        clearInterval(heartbeat);
+                        heartbeat = null;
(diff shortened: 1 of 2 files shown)


Modules/@babylonjs/react-native/EngineHook.ts
```diff
@@ -66,7 +66,48 @@ export function useEngine(): Engine | undefined {
...
+                    }
+                }
+
+                const originalStopRenderLoop: (...args: any[]) => void = engine.stopRenderLoop;
+                engine.stopRenderLoop = function(...args: any[]): void {
+                    if (--renderLoopCount == 0) {
+                        startHeartbeat();
+                    }
+                    originalStopRenderLoop.apply(this, args);
+                }
```
Alex-MSFT, line 102: So just to make sure I understand this correctly.  We stop our heartbeat task whenever the renderloop starts running, then start it back up whenever the loop goes to sleep just to keep the the looper active?

ryantrem: That's right, because when the render loop is running, it is constantly calling `requestAnimationFrame` which goes into native code, and the animation frame callback is invoked through the arcana task scheduler, which explicitly wakes up the run loop. So that ensures the looper (work queue) gets drained, though even still as mentioned above this is probably causing all promise-based async operations to be "throttled." So we almost certainly need a better fix than this, but this should at least unblock usage.

PR #66, opened 2020-08-05
Remove custom JSCRuntime

ryantrem:
This change removes our custom JSCRuntime.cpp. Two things enable this:
1. The recent upgrade to RN 0.63.1, which includes the ArrayBuffer fix in JSCRuntime.cpp.
1. The addition of a JavaScript polyfill for `atob` (just forwards to an NPM package called `base-64`). This is defined globally before Babylon Native's Window polyfill loads, so it skips defining a native `atob` polyfill. This is needed because the native polyfill uses JSI string related functions that don't work correctly for base 64 decoding because they assume strings are null terminated, and base 64 strings can effectively have null terminators in them. @bghgary have you logged a React Native bug on this yet? My change here should link to a bug.

The actual changes are:
1. Remove JSC references from build.gradle and CMakeLists.txt files.
1. Remove our custom JSCRuntime.h/JSCRuntime.cpp.
1. In the interop layer (both Android and iOS), use `jsi::Runtime` directly (don't try to extract the JSC global context and wrap it in our custom JSCRuntime).
1. Add a direct dependency on base-64.
1. Use base-64 very early in initialization to provide an `atob` polyfill.

While making these changes, I also realized that there was no reason to re-override `WebXRSessionManager.initializeSessionAsync` every time we create an engine instance, so I moved this out to be at the same scope as the `atob` global definition.

These changes should make it possible to use Hermes with the @babylonjs/react-native package because we now only depend on JSI (we don't go directly to JSC anywhere any more), but I haven't tested this yet.


bghgary:
> @bghgary have you logged a React Native bug on this yet? My change here should link to a bug.

No. I was hoping to just open a PR, but haven't gotten to it.

PR #77, opened 2020-09-10
Add takeSnapshot support to EngineView

Alex-MSFT:
This PR adds the ability to take a snapshot of content that is currently rendered for a given EngineView.

Important highlights of the implementation:
- Added a new callback prop "Initialized" to the EngineView which is used to set up a view handle that allows calling methods against the underlying native engine view.  Since EngineView is a FunctionComponent it does not actually support returning an object to make calls against it directly.
- Added takeSnapshot to the native EngineView which takes a direct snapshot of either the underlying SurfaceView or MTKView on iOS and Android respectively, and passes back the data as a Base64 encoded JPEG.
- Updated the sample app to optionally include taking snapshots.

### DIFF
--- a/Modules/@babylonjs/react-native/EngineView.tsx
+++ b/Modules/@babylonjs/react-native/EngineView.tsx
@@ -133,7 +133,7 @@ export const EngineView: FunctionComponent<EngineViewProps> = (props: EngineView
 
     // Handle snapshot data returned.
     const snapshotDataReturnedHandler = useCallback((event: SyntheticEvent) => {
-        let { data } = event.nativeEvent;
+        const { data } = event.nativeEvent;
         if (snapshotPromise) {
             snapshotPromise.resolve(data);
             setSnapshotPromise(undefined);
--- a/Modules/@babylonjs/react-native/EngineView.tsx
+++ b/Modules/@babylonjs/react-native/EngineView.tsx
@@ -98,34 +98,35 @@ export const EngineView: FunctionComponent<EngineViewProps> = (props: EngineView
         setFps(undefined);
     }, [props.camera, props.displayFrameRate]);
 
-    // Call initialized and include the hook to takeScreeenshot
+    // Call initialized and include the hook to takeScreenshot
     if (props.initialized) {
         props.initialized(
             {
                 takeScreenshot: () => {
                     if (!screenshotPromise) {
-                    let resolutionFunctions: { resolve: (data: string) => void, reject: () => void } | undefined;
-                    const promise = new Promise<string>((resolutionFunc, rejectionFunc) => {
-                        resolutionFunctions = {resolve: resolutionFunc, reject: rejectionFunc};
-                    });
+                        let resolutionFunctions: { resolve: (data: string) => void, reject: () => void } | undefined;
+                        const promise = new Promise<string>((resolutionFunc, rejectionFunc) => {
+                            resolutionFunctions = {resolve: resolutionFunc, reject: rejectionFunc};
+                        });
 
-                    if (resolutionFunctions) {
-                        setScreenshotPromise({ promise: promise, resolve: resolutionFunctions.resolve, reject: resolutionFunctions.reject });
+                        if (resolutionFunctions) {
+                            setScreenshotPromise({ promise: promise, resolve: resolutionFunctions.resolve, reject: resolutionFunctions.reject });
+                        }
+
+                        UIManager.dispatchViewManagerCommand(
+                            findNodeHandle(engineViewRef.current),
+                            UIManager.getViewManagerConfig("EngineView").Commands["takeSnapshot"],
+                            []);
+
+                        return promise;
                     }
 
-                    UIManager.dispatchViewManagerCommand(
-                        findNodeHandle(engineViewRef.current),
-                        UIManager.getViewManagerConfig("EngineView").Commands["takeSnapshot"],
-                        []);
-
-                    return promise;
-                 }
-
-                 return screenshotPromise.promise;
+                    return screenshotPromise.promise;
             }
         });
     }
 
+    // Handle snapshot data returned.
     const snapshotDataReturnedHandler = (event: SyntheticEvent) => {
         let { data } = event.nativeEvent;
         if (screenshotPromise) {
--- a/Modules/@babylonjs/react-native/EngineView.tsx
+++ b/Modules/@babylonjs/react-native/EngineView.tsx
@@ -39,7 +39,7 @@ export const EngineView: FunctionComponent<EngineViewProps> = (props: EngineView
     const [failedInitialization, setFailedInitialization] = useState(false);
     const [fps, setFps] = useState<number>();
     const engineViewRef = useRef<Component<NativeEngineViewProps>>(null);
-    const [snapshotPromise, setSnapshotPromise] = useState<{promise: Promise<string>, resolve: (data: string) => void, reject: () => void}>();
+    const snapshotPromise = useRef<{promise: Promise<string>, resolve: (data: string) => void}>();
 
     useEffect(() => {
         (async () => {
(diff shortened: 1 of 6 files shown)


Modules/@babylonjs/react-native/EngineView.tsx
```diff
@@ -90,10 +98,52 @@ export const EngineView: FunctionComponent<EngineViewProps> = (props: EngineView
         setFps(undefined);
     }, [props.camera, props.displayFrameRate]);
 
+    // Call onInitialized if provided, and include the callback for takeSnapshot.
+    useEffect(() => {
+        if (props.onInitialized) {
+            props.onInitialized({
+                takeSnapshot: () => {
+                    if (!snapshotPromise) {
```
ryantrem, line 106: The local `snapshotPromise` is being captured here, and this function is only updated if `onInitialized` changes, so I think it will always be undefined in this context, and you'd get overlapping calls to `takeSnapshot` if it was slow enough. I think you could fix this by adding `snapshotPromise` to the list of dependencies in this `useEffect`, though I'm not sure it's ok to manipulate state that cases the same `useEffect` to re-run.

I think another option would be to:
1. Store the `snapshotPromise` in a ref (e.g. `const snapshotPromise = useRef<{promise: Promise<string>, resolve: (data: string) => void}>();`). Then just read/write `snapshotPromise.current`.
1. Leave this `useEffect` basically as is (but reference `snapshotPromise.current` instead of `snapshotPromise`).
1. Remove `snapshotPromise` from the list of dependencies in the `snapshotDataReturnedHandler` `useCallback`.
This would also reduce the number of rerenders (since you don't really need to rerender when the snapshot promise changes, you just need the `takeSnapshot` function and `snapshotDataReturnedHandler` to operate on the right promise).

Modules/@babylonjs/react-native/EngineView.tsx
```diff
@@ -90,10 +98,52 @@ export const EngineView: FunctionComponent<EngineViewProps> = (props: EngineView
...
+
+                    return snapshotPromise.promise;
+                }
+            });
+        }
+    }, [props.onInitialized]);
+
+    // Handle snapshot data returned.
+    const snapshotDataReturnedHandler = useCallback((event: SyntheticEvent) => {
+        let { data } = event.nativeEvent;
```
ryantrem, line 136:
```suggestion
        const { data } = event.nativeEvent;
```

Modules/@babylonjs/react-native/android/src/main/java/com/reactlibrary/EngineView.java
```diff
@@ -37,4 +50,41 @@ public boolean onTouch(View view, MotionEvent motionEvent) {
...
+
+        // Offload the snapshot worker to a helper thread.
+        final HandlerThread helperThread = new HandlerThread("ScreenCapture",-1);
+        helperThread.start();
+        final Handler helperThreadHandler = new Handler(helperThread.getLooper());
+
+        // Request the pixel copy.
+        PixelCopy.request(this, bitmap, (copyResult) ->  {
+            // If the pixel copy was a success then convert the image to a base 64 encoded jpeg and fire the event.
+            String encoded ="";
```
ryantrem, line 76:
```suggestion
            String encoded = "";
```

Modules/@babylonjs/react-native/android/src/main/java/com/reactlibrary/EngineViewManager.java
```diff
@@ -28,4 +31,30 @@ public void onDropViewInstance(@NonNull EngineView view) {
...
+        );
+    }
+
+    @Override
+    public void receiveCommand(final EngineView view, String commandId, ReadableArray args) {
+        // This will be called whenever a command is sent from react-native.
+        switch (commandId) {
+            case COMMAND_TAKE_SNAPSHOT_NAME:
+                view.takeSnapshot();
+                break;
```
ryantrem, line 49: Should we throw in the `default` case?

Modules/@babylonjs/react-native/ios/EngineViewManager.mm
```diff
@@ -47,6 +51,28 @@ - (void)touchesCancelled:(NSSet<UITouch*>*)touches withEvent:(UIEvent*)event {
...
+        UIGraphicsBeginImageContextWithOptions(self.bounds.size, YES /* opaque */, 0.0f);
+        
+        // Draw the current state of the view into the graphics context.
+        [self drawViewHierarchyInRect:self.bounds afterScreenUpdates:NO];
+        
+        // Grab the image from the graphics context, and convert into a base64 encoded JPG.
+        UIImage *capturedImage = UIGraphicsGetImageFromCurrentImageContext();
+        UIGraphicsEndImageContext();
+        NSData *jpgData = UIImageJPEGRepresentation(capturedImage, 1.0f);
+        NSString *encodedData = [jpgData base64EncodedStringWithOptions:0];
```
ryantrem, line 67:
```suggestion
        UIImage* capturedImage = UIGraphicsGetImageFromCurrentImageContext();
        UIGraphicsEndImageContext();
        NSData* jpgData = UIImageJPEGRepresentation(capturedImage, 1.0f);
        NSString* encodedData = [jpgData base64EncodedStringWithOptions:0];
```

Modules/@babylonjs/react-native/ios/EngineViewManager.mm
```diff
@@ -59,6 +85,19 @@ @implementation EngineViewManager {
...
+RCT_EXPORT_METHOD(takeSnapshot:(nonnull NSNumber*) reactTag) {
+    // Marshal the takeSnapshot call to the appropriate EngineView.
+    [self.bridge.uiManager addUIBlock:^(RCTUIManager *uiManager, NSDictionary<NSNumber *,UIView *> *viewRegistry) {
+        EngineView *view = (EngineView *)viewRegistry[reactTag];
+        if (!view || ![view isKindOfClass:[EngineView class]]) {
+            return;
+        }
+        [view takeSnapshot];
+    }];
+}
```
ryantrem, line 99:
```suggestion
RCT_EXPORT_METHOD(takeSnapshot:(nonnull NSNumber*) reactTag) {
    // Marshal the takeSnapshot call to the appropriate EngineView.
    [self.bridge.uiManager addUIBlock:^(RCTUIManager *uiManager, NSDictionary<NSNumber*,UIView*>* viewRegistry) {
        EngineView* view = (EngineView*)viewRegistry[reactTag];
        if (!view || ![view isKindOfClass:[EngineView class]]) {
            return;
        }
        [view takeSnapshot];
    }];
}
```

Modules/@babylonjs/react-native/EngineView.tsx
Alex-MSFT: Thanks I missed this when I converted this to a useEffect.  I like your suggestion more so I will do that.

Modules/@babylonjs/react-native/android/src/main/java/com/reactlibrary/EngineViewManager.java
Alex-MSFT: Examples I've seen do not do this, but I will throw this in there for safety

Modules/@babylonjs/react-native/EngineView.tsx
```diff
@@ -90,10 +98,52 @@ export const EngineView: FunctionComponent<EngineViewProps> = (props: EngineView
...
+
+                        UIManager.dispatchViewManagerCommand(
+                            findNodeHandle(engineViewRef.current),
+                            "takeSnapshot",
+                            []);
+
+                        return promise;
+                    }
+
+                    return snapshotPromise.promise;
```
ryantrem, line 128: Isn't this broken? Doesn't it need to be `snapshotPromise.current.promise`?

Alex-MSFT: Yes good catch, fixed.  I guess it just returned undefined so it didn't actually error out, but definitely incorrect.

ryantrem (approved):

PR #81, opened 2020-09-15
Update the "heartbeat" workaround for promise continuations

ryantrem:
To work around #60 (promise continuations not executing), we have this "heartbeat" mechanism that just calls `setInterval` which seems to have the side effect of causing promise continuations that are ready to run to actually run (we still don't know the root cause). However, this workaround has two problems specifically on Android release builds:
1. Stopping the heartbeat timer after the render loop starts running is not good enough - promise continuations still get hung up.
1. Using `setInterval` somehow interferes with UI responsiveness. It seems like there is some work queue somewhere that is getting filled up, but at the same time tapping the screen (for touch interactions) *is* still responsive (but tapping UI buttons is not).

There is definitely still more investigation that needs to be done here, but I'm updating the work around to work on Android release builds as well by:
1. Continuing the heartbeat for the lifetime of the engine.
1. Switching from `setInterval` to `setTimeout` to prevent the possibility of interval callbacks piling up.

### DIFF
--- a/Modules/@babylonjs/react-native/EngineHook.ts
+++ b/Modules/@babylonjs/react-native/EngineHook.ts
@@ -76,9 +76,8 @@ export function useEngine(): Engine | undefined {
         })();
 
         // NOTE: This is a workaround for https://github.com/BabylonJS/BabylonReactNative/issues/60
-        let continueHeartbeat = true;
         function heartbeat() {
-            if (continueHeartbeat) {
+            if (!disposed) {
                 setTimeout(heartbeat, 10);
             }
         }
@@ -92,7 +91,6 @@ export function useEngine(): Engine | undefined {
                 }
                 return undefined;
             });
-            continueHeartbeat = false;
         };
     }, []);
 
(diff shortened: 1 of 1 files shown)


Alex-MSFT (approved):

Modules/@babylonjs/react-native/EngineHook.ts
```diff
@@ -71,59 +71,28 @@ export function useEngine(): Engine | undefined {
...
-                    stopHeartbeat();
-                };
-
-                setEngine(engine);
+                setEngine(new NativeEngine());
             }
         })();
 
+        // NOTE: This is a workaround for https://github.com/BabylonJS/BabylonReactNative/issues/60
+        let continueHeartbeat = true;
```
bghgary, line 79: can you use disposed for this?

bghgary (approved):

Modules/@babylonjs/react-native/EngineHook.ts
ryantrem: Hah yes.

PR #89, opened 2020-10-03
Workaround for promise continuations

bghgary:
Better fix for #60. We're asking React Native folks to see if they can confirm our theory.


Modules/@babylonjs/react-native/shared/SetTimeout.h
```diff
@@ -0,0 +1,15 @@
...
+
+#include <jsi/jsi.h>
+
+// See https://github.com/BabylonJS/BabylonReactNative/issues/60 for original issue.
+// This is a work around for what appears to be custom handling of promise continuations and
+// asynchronous functions in React Native (e.g. setTimeout, requestAnimationFrame, etc.)
+// Without a call to one of these asynchronous functions, the promise continuation is not
+// triggered. This creates a no-op setTimeout function that is called in the Babylon Native
+// JsRuntime dispatch function to poke the promise continuation execution.
+inline std::shared_ptr<facebook::jsi::Function> GetSetTimeout(facebook::jsi::Runtime& rt)
```
ryantrem, line 11: Since we are never going to actually use this as a regular `setTimeout` call (e.g. we will always pass in a zero duration and an empty callback), I'd be inclined to make the `setTimeout` part an implementation detail of what we're actually trying to achieve here - getting pending promise continuations to execute. Given that, maybe this should be more like:
`inline std::function<void()> GetPendingPromiseContinuationExecutor(facebook::jsi::Runtime& rt)`
I think this would help clarify the purpose of this code, and also simplify the call sites.

bghgary: I did this originally, but had trouble with the move semantics. I'll see if I can get it to work properly.

bghgary: I did something different. See if you're okay with it now.

ryantrem (approved):

PR #92, opened 2020-10-08
Update to latest Babylon JS/Native and fix iOS flushedQueue issue

ryantrem:
Directly calling `flushedQueue` on iOS has some side effects that I didn't notice in my first test. With some UI transitions, the internal UI state becomes broken. I'm not sure what is going on, but see the comments in the code for more details. We have an active thread on Discord with some Facebook React Native folks, so we'll wait and see if they have a better solution.

I also removed the call to reinitialize graphics right after initialization on Android as this no longer seems to be an issue.

Finally, I'm updating both Babylon.js and Babylon Native to the latest. This brings in Babylon.js changes that expose the XR tracking state, and also bring in the Babylon Native XR render loop fix.

### DIFF
--- a/Modules/@babylonjs/react-native/shared/Shared.h
+++ b/Modules/@babylonjs/react-native/shared/Shared.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <napi/napi.h>
+#include <jsi/jsi.h>
 
 // See https://github.com/BabylonJS/BabylonReactNative/issues/60 for original issue.
 // This is a work around to poke the React Native message queue to run setImmediate callbacks.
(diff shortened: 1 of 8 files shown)


Modules/@babylonjs/react-native/shared/Shared.h
```diff
@@ -16,3 +17,14 @@ inline Napi::FunctionReference GetFlushedQueue(Napi::Env env)
...
+
+// On iOS, directly calling flushedQueue breaks some kind of internal UI state and we start getting errors like:
+// *** Assertion failure in -[RCTNativeAnimatedNodesManager disconnectAnimatedNodes:childTag:](), node_modules/react-native/Libraries/NativeAnimation/RCTNativeAnimatedNodesManager.m:138
+// [native] Exception thrown while executing UI block: 'parentNode' is a required parameter
+// However, calling setTimeout eventually also calls flushedQueue, but doesn't result in the above error,
+// so we'll go this route until we (hopefully) get a better resolution from the React Native team at Facebook.
+inline std::shared_ptr<facebook::jsi::Function> GetSetTimeout(facebook::jsi::Runtime& rt)
+{
+    auto code{std::make_shared<const facebook::jsi::StringBuffer>("() => setTimeout(() => {})")};
+    return std::make_shared<facebook::jsi::Function>(rt.evaluateJavaScript(std::move(code), "").asObject(rt).asFunction(rt));
```
bghgary, line 29: Can/should you use napi here instead of jsi?

bghgary (approved):

Modules/@babylonjs/react-native/shared/Shared.h
ryantrem: I would assume not? Why bring in an extra layer where we don't need the napi abstraction?

bghgary: I used napi in the other function..

ryantrem: We can revisit this if you want to talk about it more, or we can wait and see if we get a better solution in the Discord thread.

PR #94, opened 2020-10-12
Use Babylon Native's XMLHttpRequest

ryantrem:
React Native's `XMLHttpRequest` uses too much memory and is too slow for use with Babylon.js. This change brings in Babylon Native's `XMLHttpRequest`, but does not replace the globally scoped one provided by React Native. Instead, it exposes it from the `_native` object, which `WebRequest` will specifically look for in @babylonjs/core@4.2.0-beta.12 or newer.
- Update Babylon.js to 4.2.0-beta.12.
- Update Babylon Native to latest (for corresponding `XMLHttpRequest` related changes).
- Add `XMLHttpRequest` to the Android and iOS build.
- Initialize `XMLHttpRequest` in the Android and iOS interop layers.
- Add `XMLHttpRequest` related libs to the pod file and package build validation.

Other unrelated changes included in this PR:
- Updating some GitHub build tasks to the latest to see if it resolves some confusing build warnings.
- Update iOS test app project files to use iOS 12 to match what the Babylon React Native podspec requires (otherwise I get errors when building Release in XCode 12).
- Revert to the old workaround for the promise continuation issue (it turns out the new approach doesn't work on iOS in Release builds).
- Add `@react-native-community/cli` as a dev dependency to the test apps as it seems we can use it to convert a Hermes profile to CDT profiler format and view it in Chrome (still trying to get this to work).

### DIFF
--- a/Modules/@babylonjs/react-native/ios/BabylonNative.cpp
+++ b/Modules/@babylonjs/react-native/ios/BabylonNative.cpp
@@ -75,6 +75,8 @@ namespace Babylon
         m_impl->m_graphics->AddToJavaScript(m_impl->env);
 
         Polyfills::Window::Initialize(m_impl->env);
+        // NOTE: React Native's XMLHttpRequest is slow and allocates a lot of memory. This does not override
+        // React Native's implementation, but rather adds a second one scoped to Babylon and used by WebRequest.ts.
         Polyfills::XMLHttpRequest::Initialize(m_impl->env);
 
         Plugins::NativeWindow::Initialize(m_impl->env, windowPtr, width, height);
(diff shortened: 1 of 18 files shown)


Modules/@babylonjs/react-native/ios/BabylonNative.cpp
```diff
@@ -72,6 +75,7 @@ namespace Babylon
         m_impl->m_graphics->AddToJavaScript(m_impl->env);
 
         Polyfills::Window::Initialize(m_impl->env);
+        Polyfills::XMLHttpRequest::Initialize(m_impl->env);
```
bghgary, line 78: Maybe a comment here indicating why we have a separate XMLHttpRequest on top of RN's?

bghgary (approved):

PR #95, opened 2020-10-13
Validate Babylon.js version

ryantrem:
This change brings in the `semver` lib and validates the Babylon.js version (from `Engine.Version`) against the dependency listed in `package.json`. If the version of @babylonjs/core is too old, it logs an error with `console.error`, which shows up in the console window as an error and also as a red box error in the host app. I think the only downside of this is the up to 86k added by the `semver` lib.


Alex-MSFT (approved):

ryantrem:
Folks are still getting bitten by having a Babylon.js that is too old for Babylon React Native (@swese44 recently with #148 for example). The NPM warning is too easy to ignore, and then the broken behavior is hard to diagnose. Given this, I'm moving this PR out of draft.

PR #107, opened 2020-10-28
Update to latest Babylon Native with render threading

ryantrem:
This change updates to the latest BabylonNative submodule, and makes changes to work with the render threading updates (https://github.com/BabylonJS/BabylonNative/pull/451).

### DIFF
--- a/Modules/@babylonjs/react-native/ios/BabylonNative.cpp
+++ b/Modules/@babylonjs/react-native/ios/BabylonNative.cpp
@@ -2,7 +2,6 @@
 
 #include <Babylon/Graphics.h>
 #include <Babylon/JsRuntime.h>
-#include <Babylon/Plugins/NativeWindow.h>
 #include <Babylon/Plugins/NativeEngine.h>
 #include <Babylon/Plugins/NativeInput.h>
 #include <Babylon/Plugins/NativeXr.h>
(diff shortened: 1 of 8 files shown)


Modules/@babylonjs/react-native/ios/BabylonNative.cpp
```diff
@@ -57,8 +56,7 @@ namespace Babylon
         // React Native's implementation, but rather adds a second one scoped to Babylon and used by WebRequest.ts.
         Polyfills::XMLHttpRequest::Initialize(m_impl->env);
 
-        Plugins::NativeWindow::Initialize(m_impl->env, windowPtr, width, height);
-        Plugins::NativeEngine::Initialize(m_impl->env);
+        Plugins::NativeEngine::Initialize(m_impl->env, true);
```
syntheticmagus, line 59: nit: `true` is the default value for that argument; you can specify it explicitly, but you don't need to.

syntheticmagus (approved):

syntheticmagus:
Technically, you should probably add something to your JavaScript to [await the graphics promise](https://github.com/BabylonJS/BabylonNative/blob/master/Apps/Playground/Scripts/experience.js#L54), but it won't really matter or be necessary until you start actually rendering multithreaded. If `NativeEngine` is auto-rendering, that promise is completed synchronously.

Modules/@babylonjs/react-native/ios/BabylonNative.cpp
ryantrem: Ok, I think I'd rather be explicit with this arg.

ryantrem:
> Technically, you should probably add something to your JavaScript to [await the graphics promise](https://github.com/BabylonJS/BabylonNative/blob/master/Apps/Playground/Scripts/experience.js#L54), but it won't really matter or be necessary until you start actually rendering multithreaded. If `NativeEngine` is auto-rendering, that promise is completed synchronously.

Fixed so we don't forget about it later.

PR #112, opened 2020-11-12
Get dev mode reload working

### LINKED ISSUE #84: Reloading the JS engine causes a crash
This used to work, so it is a regression, but on both iOS and Android if you reload the JS engine (either hit 'R' in the react native console window, or do a `Reload` from the dev menu inside the app), the app crashes. Fast refresh works fine, but a full reload causes the crash.

ryantrem:
This change includes a bunch of fixes for React Native dev mode reload (which reloads the JS engine).
- Update to the latest BabylonNative submodule which includes a fix to actually delete the napi instance on `Napi::Detach`.
- Set the current `NativeEngine` instance on the `_native` object so we can get it on the native side and manually dispose it (since during a reload the `EngineHook`'s cleanup function will not be called, but we still need to release these resources).
- Have the native interop code grab the current `NativeEngine` instance off of `_native` during teardown and call `dispose` on it.
- Prevent the JS dispatcher from actually doing anything if we are in the process of shutting down the JS engine.

These changes are a little icky and there is a fair bit of duplication across Android and iOS. I have a bigger cleanup planned for this code that will consolidate a lot of the Android and iOS code, but that will happen later.

Fixes #84

### DIFF
--- a/Modules/@babylonjs/react-native/BabylonModule.ts
+++ b/Modules/@babylonjs/react-native/BabylonModule.ts
@@ -1,6 +1,31 @@
 import { NativeModules } from 'react-native';
+import { NativeEngine } from '@babylonjs/core';
 
-export const BabylonModule: {
+// This global object is part of Babylon Native.
+declare const _native: {
+    graphicsInitializationPromise: Promise<void>;
+    engineInstance: NativeEngine;
+}
+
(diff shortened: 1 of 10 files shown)


bghgary (approved):
Looks good

Issue #19, opened 2020-06-01
App crashes during fast refresh if in AR mode


Alex-MSFT:
If a hot reload is triggered while in AR mode the app crashes.  If you exit AR mode first hot reload works as expected.  With the below traces (probably because the JS Engine is getting torn down).

We should gracefully handle this failure and exit AR mode.

2020-06-01 10:24:25.209 907-959/com.playground A/libc: E:/BabylonReactNative/Apps/Playground/node_modules/react-native-babylon/submodules/BabylonNative/Dependencies/napi/source/js_native_api_JavaScriptCore.cc:114: JSObjectRef (anonymous namespace)::ToJSObject(napi_env, const napi_value): assertion "value == nullptr || JSValueIsObject(env->context, reinterpret_cast<JSValueRef>(value))" failed
2020-06-01 10:24:25.209 907-959/com.playground A/libc: Fatal signal 6 (SIGABRT), code -6 (SI_TKILL) in tid 959 (mqt_js), pid 907 (com.playground)

ryantrem:
I think this was for a full JS engine reload (e.g. press R in the console window) and not actual hot reload/fast refresh. This is probably blocked by #84 now.

bghgary:
Blocked by #84

ryantrem:
#84 has been fixed, so this is no longer blocked. I took a quick look and one problem I saw right away is that the XR session is not synchronously disposed when the engine is disposed (which is not the behavior I was expecting). This is because `NativeXR.cpp` does the following:

```c++
m_xr.Dispatch([this]() {
    m_xr.EndSession();
    ...
```

Basically makes a fire-and-forget async call to `EndSession`. This means the session remains alive even after the engine has been disposed. I tried just removing the `Dispatch`, but then we get a crash in bgfx. We need to dig in a bit deeper to understand the order of operations with ending the XR session, queued frame rendering, etc.

bghgary:
Closing for now. Reopen if we want to fix this.

PR #122, opened 2020-11-25
Update to latest BabylonNative with tracking states exposed

CoPrez:
With this change I'm updating BabylonNative to the latest from master which exposes the tracking states. Along with updating BabylonNative I'm also adding in some improvements to the Playground app. I've updated the Playground app to display a message about what state XR tracking is in. Additionally I am checking in some VSCode settings that add debug targets and make deployment/testing easier within VSCode.

With these changes I performed the following manual tests on Android and iOS for Debug and Release flavors:
1. [x] **Basic rendering** - launch the Playground app and make sure the model loaded and is rendering at 60fps.
1. [x] **Animation** - make sure the loaded model is animating.
1. [x] **Input handling** - swipe across the display and make sure the model rotates around the y-axis.
1. [x] **Display rotation** - rotate the device 90 degrees and make sure the view rotates and renders correctly.
1. [x] **View replacement** - tap the *Toggle EngineView* button twice to replace the render target view.
1. [x] **Engine dispose** - tap the *Toggle EngineScreen* button twice to dispose and re-instantiate the Babylon engine.
1. [x] **Suspend/resume** - switch to a different app and then back to the Playground and make sure it is still rendering correctly.
1. [x] **Dev mode reload** - in the Metro server console window, press the `R` key on the keyboard to reload the JS engine and make sure rendering restarts successfully.
1. [x] **XR mode** - tap the *Start XR* button and make sure XR mode is working.
1. [x] **XR display rotation** - rotate the device 90 degrees and make sure the view rotates and renders correctly.
1. [x] **XR view replacement** - tap the *Toggle EngineView* button twice to replace the render target view.
1. [x] **XR suspend/resume** - switch to a different app and then back to the Playground and make sure it is still rendering correctly.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,38 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Debug Android",
+            "cwd": "${workspaceFolder}/Apps/Playground",
+            "type": "reactnativedirect",
+            "request": "launch",
+            "platform": "android"
+        },
+        {
+            "name": "Debug iOS on Device",
+            "cwd": "${workspaceFolder}/Apps/Playground",
+            "type": "reactnativedirect",
+            "request": "launch",
+            "platform": "ios",
+            "port": 9221,
+            "target": "device"
+        },
+        {
+            "name": "Attach to Android",
+            "cwd": "${workspaceFolder}/Apps/Playground",
+            "type": "reactnativedirect",
+            "request": "attach"
+        },
+        {
+            "name": "Attach to iOS on Device",
+            "cwd": "${workspaceFolder}/Apps/Playground",
+            "type": "reactnativedirect",
+            "request": "attach",
+            "platform": "ios",
+            "port": 9221
+        },
+    ]
+}
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,14 +5,14 @@
     "version": "0.2.0",
     "configurations": [
         {
-            "name": "Debug Android",
+            "name": "Build & Debug Android",
             "cwd": "${workspaceFolder}/Apps/Playground",
             "type": "reactnativedirect",
             "request": "launch",
             "platform": "android"
         },
         {
-            "name": "Debug iOS on Device",
+            "name": "Build & Debug iOS on Device",
             "cwd": "${workspaceFolder}/Apps/Playground",
             "type": "reactnativedirect",
             "request": "launch",
--- a/Apps/Playground/App.tsx
+++ b/Apps/Playground/App.tsx
@@ -75,23 +75,8 @@ const EngineScreen: FunctionComponent<ViewProps> = (props: ViewProps) => {
     }
   }, [rootNode, scale]);
 
-  const trackingStateToString = (trackingState: WebXRTrackingState | undefined) : string =>
-  {
-    if (trackingState == undefined) {
-      return ""
-    }
-
-    switch(trackingState) {
-      case WebXRTrackingState.NOT_TRACKING:
-        return "Not Tracking";
-        break;
-      case WebXRTrackingState.TRACKING:
-        return "Tracking";
-        break;
-      case WebXRTrackingState.TRACKING_LOST:
-        return "Tracking Lost";
-        break;
-    }
+  const trackingStateToString = (trackingState: WebXRTrackingState | undefined) : string => {
+    return trackingState === undefined ? "" : WebXRTrackingState[trackingState];
   }
 
   const onToggleXr = useCallback(() => {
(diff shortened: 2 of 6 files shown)


Alex-MSFT:
> **Note -** I'm currently investigating why **Dev mode reload** is not passing with my changes

Dev mode reload is a known issue #19, not a regression from this change

Apps/Playground/App.tsx
```diff
@@ -74,16 +75,42 @@ const EngineScreen: FunctionComponent<ViewProps> = (props: ViewProps) => {
     }
   }, [rootNode, scale]);
 
+  const trackingStateToString = (trackingState: WebXRTrackingState | undefined) : string =>
+  {
```
Alex-MSFT, line 79:
```suggestion
  const trackingStateToString = (trackingState: WebXRTrackingState | undefined) : string => {
```

Apps/Playground/App.tsx
```diff
@@ -74,16 +75,42 @@ const EngineScreen: FunctionComponent<ViewProps> = (props: ViewProps) => {
...
         setXrSession(undefined);
+        setTrackingState(undefined);
       } else {
         if (rootNode !== undefined && scene !== undefined) {
           const xr = await scene.createDefaultXRExperienceAsync({ disableDefaultUI: true, disableTeleportation: true })
           const session = await xr.baseExperience.enterXRAsync("immersive-ar", "unbounded", xr.renderTarget);
           setXrSession(session);
+
+          setTrackingState(xr.baseExperience.camera.trackingState);
+          xr.baseExperience.camera.onTrackingStateChanged.add((newTrackingState) =>{
```
Alex-MSFT, line 110:
```suggestion
          xr.baseExperience.camera.onTrackingStateChanged.add((newTrackingState) => {
```

Apps/Playground/App.tsx
```diff
@@ -74,16 +75,42 @@ const EngineScreen: FunctionComponent<ViewProps> = (props: ViewProps) => {
...
+            setTrackingState(newTrackingState);
+          })
```
Alex-MSFT, line 112:
```suggestion
          });
```

Apps/Playground/App.tsx
```diff
@@ -74,16 +75,42 @@ const EngineScreen: FunctionComponent<ViewProps> = (props: ViewProps) => {
...
+        return "Not Tracking";
+        break;
+      case WebXRTrackingState.TRACKING:
+        return "Tracking";
+        break;
+      case WebXRTrackingState.TRACKING_LOST:
+        return "Tracking Lost";
+        break;
+    }
+  }
```
Alex-MSFT, line 95:
```suggestion
  };
```

Apps/Playground/App.tsx
```diff
@@ -74,16 +75,42 @@ const EngineScreen: FunctionComponent<ViewProps> = (props: ViewProps) => {
     }
   }, [rootNode, scale]);
 
+  const trackingStateToString = (trackingState: WebXRTrackingState | undefined) : string =>
+  {
+    if (trackingState == undefined) {
+      return ""
+    }
+
+    switch(trackingState) {
```
Alex-MSFT, line 84:
```suggestion
    switch (trackingState) {
```

Alex-MSFT (approved):

ryantrem:
I discovered late last week that dev mode reload is broken when a glb has been loaded. Try commenting out the glb load in App.tsx.

Apps/Playground/App.tsx
```diff
@@ -74,16 +75,42 @@ const EngineScreen: FunctionComponent<ViewProps> = (props: ViewProps) => {
...
+      case WebXRTrackingState.NOT_TRACKING:
+        return "Not Tracking";
+        break;
+      case WebXRTrackingState.TRACKING:
+        return "Tracking";
+        break;
+      case WebXRTrackingState.TRACKING_LOST:
+        return "Tracking Lost";
+        break;
+    }
```
ryantrem, line 94: I would just do this:
```suggestion
    return trackingState === undefined ? "" : WebXRTrackingState[trackingState];
```
That will give you strings like "NOT_TRACKING" but I think that is fine. Your call, but I think the simpler code that results in slightly less polished state string is a bit better.

.vscode/launch.json
```diff
@@ -0,0 +1,38 @@
...
+    "configurations": [
+        {
+            "name": "Build & Debug Android",
+            "cwd": "${workspaceFolder}/Apps/Playground",
+            "type": "reactnativedirect",
+            "request": "launch",
+            "platform": "android"
+        },
+        {
+            "name": "Build & Debug iOS on Device",
```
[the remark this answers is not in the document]
CoPrez: Android will use the first device it finds whether it is an emulator or a physical device. iOS uses a different pipeline for targeting the simulator versus targeting a physical device so it's explicit on iOS.

Apps/Playground/App.tsx
CoPrez: I like that a lot! I wasn't aware of that syntax.

ryantrem (approved):

PR #135, opened 2020-12-19
Converge react native integration layers across platforms

### LINKED ISSUE #24: Replace usages of CompletableFuture
The Android implementation currently uses CompletableFuture, but this is not available in any API level below 24, which is higher than the minSdkVersion that we want to support. As such, we need to remove usages of CompletableFuture and replace them with some alternative.

ryantrem:
This change significantly overhauls the react native integration layers. The goal is to reduce duplication across platforms and simplify the code. The main shared integration code is in `BabylonNative.cpp`. This is similar to what we had for both Android and iOS, but is platform agnostic. Android and iOS have relatively thin platform specific layers on top of this. Additionally, the class defined in `BabylonNative.cpp` is now a JSI `HostObject` which is injected into the JavaScript engine so that JS code can call directly into it in a platform agnostic way (so we don't have to expose native functionality to JS through platform specific legacy bridge based modules). This object exposes a JS promise that JS Babylon React Native initialization code can await, which allowed me to remove all the platform specific native promise/callback related code. I was also able to remove all traces of the old usages of loopers, since we fully rely on the React Native `CallInvoker` to get onto the JS thread.

NOTE: This change requires a change in BabylonNative that is not merged yet, so I'll need to update that once https://github.com/BabylonJS/BabylonNative/pull/550 is merged.

This also resolves #24


Modules/@babylonjs/react-native/shared/BabylonNative.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <jsi/jsi.h>
+#include <ReactCommon/CallInvoker.h>
+
+namespace Babylon
+{
+    void Initialize(facebook::jsi::Runtime& jsiRuntime, std::shared_ptr<facebook::react::CallInvoker> jsCallInvoker);
+    void Deinitialize();
```
chrisfromwork, line 9: how important do we feel it is for all platforms to use these function definitions? Right now UWP does not have a call invoker. It also needs to specify window types during update view. Could these functions be templated/generics?

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
```diff
@@ -0,0 +1,230 @@
...
+        std::unique_ptr<Graphics> m_graphics{};
+        Plugins::NativeInput* m_nativeInput{};
+
+        std::function<void()> m_disposeEngine{};
+    };
+
+    namespace
+    {
+        constexpr auto JS_INSTANCE_NAME{ "BabylonNative" };
+        std::weak_ptr<ReactNativeModule> g_nativeModule{};
```
chrisfromwork, line 182: >g_nativeModule [](start = 41, length = 14)

should this be a map with jsiruntimes as keys? I don't know that we'll ever live in a world where we see multiple jsiruntimes in a single application life cycle, but it seems like a jsiruntime to native module mapping may make more sense than a global/static variable

Modules/@babylonjs/react-native/shared/BabylonNative.h
chrisfromwork: do we specifically need to hand around this call invoker or could we just hand around the dispatch function? Handing around the dispatch function and allowing the window type to be set when updating the view seems like it would unblock UWP scenarios

---
In reply to: [546838824](https://github.com/BabylonJS/BabylonReactNative/pull/135#discussion_r546838824) [](ancestors = 546838824)

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
```diff
@@ -0,0 +1,230 @@
...
+                    m_graphics->AddToJavaScript(m_env);
+                    Plugins::NativeEngine::Initialize(m_env, true);
+                    m_resolveInitPromise();
+                }
+                else
+                {
+                    m_graphics->UpdateWindow(windowPtr);
+                    m_graphics->UpdateSize(width, height);
+                    m_graphics->EnableRendering();
+                }
```
chrisfromwork, line 92: for UWP we will need a mechanism to allow for calling graphics render functionality from the UI thread. If the thought is that we will have a different BabylonNative setup for UWP, that is fine too

Modules/@babylonjs/react-native/BabylonModule.ts
```diff
@@ -1,33 +1,72 @@
...
+};
 
-const NativeBabylonModule: {
-    initialize(): Promise<boolean>;
-    whenInitialized(): Promise<boolean>;
-    reset(): Promise<boolean>;
+// This legacy React Native module is created by Babylon React Native, and is only used to bootstrap the JSI object creation.
+// This will likely be removed when the BabylonNative global object is eventually converted to a TurboModule.
+const BabylonModule: {
+    initialize(): Promise<void>;
```
chrisfromwork, line 27: >initialize [](start = 4, length = 10)

is there a reason we aren't moving this initialization call to BabylonNative now?

Modules/@babylonjs/react-native/BabylonModule.ts
```diff
@@ -1,33 +1,72 @@
...
 declare const _native: {
     whenGraphicsReady: () => Promise<void>;
-    engineInstance: NativeEngine;
-}
+};
+
+// This JSI-based global object is owned by Babylon React Native.
+// This will likely be converted to a TurboModule when they are fully supported.
+declare const BabylonNative: {
+    readonly initializationPromise: Promise<void>;
```
chrisfromwork, line 19: >initializationPromise [](start = 13, length = 21)

does this need to be a property or could it be a function? I have not seen great examples in react-native-windows for handing back non-value/non-object types as properties

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        void SetPointerPosition(uint32_t pointerId, uint32_t x, uint32_t y)
+        {
+            m_nativeInput->PointerMove(pointerId, x, y);
+        }
+
+        jsi::Value get(jsi::Runtime& runtime, const jsi::PropNameID& prop) override
+        {
+            if (prop.utf8(runtime) == "initializationPromise")
```
bghgary, line 130: Probably should store `prop.utf8(runtime)` in a variable or it will reallocate the resulting string multiple times.

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
```diff
@@ -0,0 +1,230 @@
...
+                        {
+                            auto engineInstance = engineInstanceValue->getObject(rt);
+                            engineInstance.getPropertyAsFunction(rt, "dispose").callWithThis(rt, engineInstance);
+                        };
+                    }
+                    return {};
+                });
+            }
+            
+            return jsi::Value::undefined();
```
bghgary, line 162: nit
```suggestion
            return {};
```

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+    namespace
+    {
+        constexpr auto JS_INSTANCE_NAME{ "BabylonNative" };
+        std::weak_ptr<ReactNativeModule> g_nativeModule{};
+    }
+
+    void Initialize(facebook::jsi::Runtime& jsiRuntime, std::shared_ptr<facebook::react::CallInvoker> jsCallInvoker)
+    {
+        if (jsiRuntime.global().getProperty(jsiRuntime, JS_INSTANCE_NAME).isUndefined())
```
bghgary, line 187: nit
```suggestion
        if (jsiRuntime.global().hasProperty(jsiRuntime, JS_INSTANCE_NAME))
```

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+    void UpdateView(void* windowPtr, size_t width, size_t height)
+    {
+        if (auto nativeModule = g_nativeModule.lock())
+        {
+            nativeModule->UpdateView(windowPtr, width, height);
+        }
+        else
+        {
+            throw std::runtime_error { "UpdateView must not be called before Initialize." };
```
bghgary, line 211: nit
```suggestion
            throw std::runtime_error{ "UpdateView must not be called before Initialize." };
```

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
```diff
@@ -0,0 +1,230 @@
...
+        {
+            auto nativeModule{ std::make_shared<ReactNativeModule>(jsiRuntime, jsCallInvoker) };
+            jsiRuntime.global().setProperty(jsiRuntime, JS_INSTANCE_NAME, jsi::Object::createFromHostObject(jsiRuntime, nativeModule));
+            g_nativeModule = nativeModule;
+        }
+    }
+
+    void Deinitialize()
+    {
+        if (auto nativeModule = g_nativeModule.lock())
```
bghgary, line 197: nit
```suggestion
        if (auto nativeModule{ g_nativeModule.lock() })
```

bghgary (commented):
Did a pass on it, but I can't say I understand everything. Maybe good to go over in person (virtually).

Modules/@babylonjs/react-native/BabylonModule.ts
```diff
@@ -1,33 +1,72 @@
...
+// This global object is owned by Babylon Native.
 declare const _native: {
     whenGraphicsReady: () => Promise<void>;
-    engineInstance: NativeEngine;
-}
+};
+
+// This JSI-based global object is owned by Babylon React Native.
+// This will likely be converted to a TurboModule when they are fully supported.
+declare const BabylonNative: {
```
chrisfromwork, line 18: >BabylonNative [](start = 14, length = 13)

 is BabylonNative supposed to be a native module or a jsi/napi constructed object? Is it fair to assume that all constants loaded via NativeModules.* are react-native native modules while constants not loaded via NativeModules.* are jsi/napi populated?

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
ryantrem: I did that in the previous implementation on Android, and ultimately I thought it made it a lot more complex to cover a scenario that I don't think we'll ever have (multiple JS engine instances existing at the same time). It's not just the map that makes it more complex, it also means the `jsi::Runtime` must be passed to every one of the C APIs that I expose. Again, to me this just seems like added complexity for yagni. If you disagree let me know and I'm happy to explore it further.

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
ryantrem: My intent is that this will be shared amongst all platforms, including UWP. I understand that the platform specific code needs to be able to determine the rendering thread and probably manually call `RenderCurrentFrame` directly, but that was just too much to try to fit into this already disruptive change. My plan was to tackle that separately after this PR.

Modules/@babylonjs/react-native/BabylonModule.ts
ryantrem: Yes. At this point in the code, the `BabylonNative` object has not been instantiated and injected into the JS execution context. This method serves as a way to "bootstrap" the initialization process and results in the `BabylonNative` instance being instantiated and injected into the JS execution context. After this is done, all other JS<==>C++ communication can go through the `BabylonNative` object.

Modules/@babylonjs/react-native/BabylonModule.ts
ryantrem: It could be a function if needed, it just didn't seem necessary. I would be surprised if this wasn't supported in react-native-windows, but if we find out its not, I will change it at that time to a function.

chrisfromwork: I talked some with react-native-windows owners and if this is supposed to be a native module, this isn't supported. They also said that setEngineInstance isn't supported either as a function for a native module in UWP. We may want to limit the native module interface to the bare minimum and populate everything else with JSI/NAPI.

Modules/@babylonjs/react-native/EngineView.tsx
```diff
@@ -145,11 +132,11 @@ export const EngineView: FunctionComponent<EngineViewProps> = (props: EngineView
         }
     }, []);
 
-    if (!failedInitialization) {
+    if (initialized !== false) {
```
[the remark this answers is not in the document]
ryantrem: `initialized` is `undefined` until the async call to `ensureInitialized` is complete. So there are three possible states - waiting for initialization, initialization succeeded, initialization failed.

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
ryantrem: This is a very coarse grained operation. In most apps, it will only happen once. Does caching really matter in this case?

Modules/@babylonjs/react-native/BabylonModule.ts
ryantrem: By `NativeModules.*`, do you mean on the JS side, such as:

```javascript
const BabylonModule: {
    initialize(): Promise<void>;
} = NativeModules.BabylonModule;
```

If that is what you mean by `NativeModules.*`, then those are instantiated by React Native itself. `BabylonNative` is a JSI `HostObject` that we manually construct and inject into the JS execution context (which I think is what you mean by jsi/napi constructed/populated). If that is all correct, then the answer to your question is yes.

In the future, we should be able to get rid of `BabylonModule`. When TurboModules are fully supported, a TurboModule is itself just a derivation of `HostObject`, so transitioning the `BabylonNative` object to a TurboModule (when fully supported) should be straight forward (and will let us remove `BabylonModule`).

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
bghgary: Nope, but it's still not a great pattern. I think we should fix it so that no one copies it.

ryantrem: Oh, I totally misunderstood what you were originally suggesting. Will fix.

Modules/@babylonjs/react-native/BabylonModule.ts
ryantrem: Let's try to sync up on this tomorrow.

Modules/@babylonjs/react-native/shared/BabylonNative.h
ryantrem: We could probably pass an `std::function` instead of a `CallInvoker`. I set it up to pass a `CallInvoker` because:
1. `CallInvoker` does extra work to get JS promise continuations to run. Without it, promise continuations can "hang" indefinitely. We saw this behavior on both iOS and Android, and I expect it is the same on UWP.
2. `CallInvoker` exposes functions to execute work on the JS thread both synchronously and asynchronously. That said, I discovered that the `CallInvoker::invokeSync` function is not actually implemented (it just throws an unsupported exception).

So again, we should be able to switch it to `std::function`, but I suspect if we don't use `CallInvoker` on UWP, we are going to have problems with JS promise continuations.

chrisfromwork: chatted offline, sounds like we are using CallInvoker on UWP, but we don't have access to it. Because we are using CallInvoker, we felt confident that handing around a std::function using the CallInvoker would still work well here

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
chrisfromwork: I don't have a strong preference. I think on the UWP side it was straight forward to initialize per jsiruntime/react context. I have not actually seen any scenarios where multiple jsiruntimes/react contexts are created/used.

Modules/@babylonjs/react-native/BabylonModule.ts
chrisfromwork: Talked offline. I was confused here thinking this was a module definition when it will be a jsi/napi object declaration. This should work fine

Modules/@babylonjs/react-native/BabylonModule.ts
chrisfromwork: Chatted offline, this makes a lot more sense now. It sounds like NativeModules.* and requireNativeComponent is following what's expected for react-native

Modules/@babylonjs/react-native/shared/BabylonNative.h
chrisfromwork: @ryantrem feel free to resolve this convesation, i'll keep it open for now while we wait on changes to pass around a std function

chrisfromwork (approved):

bghgary (approved):

PR #136, opened 2021-01-07
Bring in newest BabylonNative & Babylon.js

Alex-MSFT:
Bringing in the newest Babylon.js version, and updating BabylonNative to fix crashes when exiting XR Mode for Babylon.js versions newer than 5.0.0.

After this commit only versions of Babylon.js newer than 5.0.0-alpha.6 are compatible with BabylonReact-Native due to breaking changes between Babylon.js & BabylonNative.

### DIFF
--- a/Modules/@babylonjs/react-native/package.json
+++ b/Modules/@babylonjs/react-native/package.json
@@ -27,7 +27,7 @@
     "base-64": "^0.1.0"
   },
   "peerDependencies": {
-    "@babylonjs/core": "^4.2.0",
+    "@babylonjs/core": "^5.0.0-alpha.6",
     "react": "^16.13.1",
     "react-native": "^0.63.1",
     "react-native-permissions": "^2.1.4"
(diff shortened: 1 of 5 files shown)


ryantrem:
Make sure to update the @babylonjs/core min version in `Modules/@babylonjs/react-native/package.json` as well.

Modules/@babylonjs/react-native/package.json
```diff
@@ -27,7 +27,7 @@
     "base-64": "^0.1.0"
   },
   "peerDependencies": {
-    "@babylonjs/core": "^4.2.0",
+    "@babylonjs/core": "^5.0.0-alpha.6",
```
ryantrem, line 30: I guess it doesn't really matter that much, but didn't you say the min ver is 5.0.0-alpha.1?

ryantrem (approved):

Modules/@babylonjs/react-native/package.json
Alex-MSFT: I forgot there's actually a change needed in alpha.6 for tearing down the XR session updating the description.

PR #137, opened 2021-01-07
Add UWP support to BabylonReactNative

chrisfromwork:
This pr adds support for react-native-windows. Instructions for building the react-native-windows babylon.js package can be found at Modules\@babylonjs\react-native\windows\README.txt

## Overview:
* A CMakeLists.txt file has been added for the windows @babylonjs/react-native package that sets up BabylonNative dependencies, the react-native needed jsi project and a BabylonNative.lib that utilizes pre-existing BabylonReactNative BabylonNative initialization logic
* We've updated BabylonNative.lib to expose the ability to specify a windowTypePtr. The windowTypePtr needs to be specified to use XAML SwapChainPanels when rendering
* BabylonReactNative.dll has been added to the @babylonjs/react-native package to implement EngineView/EngineViewManager/BabylonModule native components
* The Playground app has been updated to support react-native-windows scenarios
* @react-native-community/slider has been updated to a newer version that supports react-native-windows
* @react-native-permissions has been updated to a newer version that supports react-native-windows
* react and react-native have been updated to versions that support react-native-windows
* metro.config.js for the Playground app has been updated to exclude windows files that should not be bundled/deployed

## TODOs:
- [x] Fix Slider logic/find a slider implementation that is supported on all platforms
- [x] Fix xr camera permissions logic. Permissions is not currently a supported react-native-windows native module
- [x] Refactor logic to support newer BabylonReactNative commits

## Follow up issues to file:
- [x] OpenXR errors are encountered when attempting to enter immersive vr on a desktop when no vr headset is plugged in/mixed reality portal isn't running: https://github.com/BabylonJS/BabylonNative/issues/579
- [x] Fix NativeInput, which does not appear to work correctly for UWP: https://github.com/BabylonJS/BabylonReactNative/issues/139
- [x] Stopping xr generates crashes: https://github.com/BabylonJS/BabylonReactNative/issues/140
- [x] Fix Toggle EngineView lifecycle, which fails to draw to the new swap chain panel: https://github.com/BabylonJS/BabylonReactNative/issues/140
- [x] Fix Toggle EngineScreen lifecycle, which fails to draw to the correct swap chain panel: https://github.com/BabylonJS/BabylonReactNative/issues/140
- [x] Package compiled BabylonReactNative.winmd/BabylonReactNative.dll into the distributed npm package: https://github.com/BabylonJS/BabylonReactNative/issues/141


Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineViewManager.cpp
```diff
@@ -0,0 +1,143 @@
...
+    // IViewManager
+    hstring EngineViewManager::Name() noexcept {
+        return L"EngineView";
+    }
+
+    FrameworkElement EngineViewManager::CreateView() noexcept {
+        _swapChainPanel = SwapChainPanel();
+        _swapChainPanelPtr = nullptr;
+        copy_to_abi(_swapChainPanel, _swapChainPanelPtr);
+        _swapChainPanel.SizeChanged({ this, &EngineViewManager::OnSizeChanged });
```
chrisfromwork, line 28: @tymlipari do you have suggestions for how we should be subscribing/auto unsubscribing to these events?

Apps/Playground/package.json
```diff
@@ -13,29 +15,35 @@
     "@babylonjs/core": "^5.0.0-alpha.6",
     "@babylonjs/loaders": "^5.0.0-alpha.6",
     "@babylonjs/react-native": "file:../../Modules/@babylonjs/react-native",
-    "@react-native-community/slider": "^2.0.9",
+    "@react-native-community/slider": "4.0.0-rc.2",
     "logkitty": "^0.7.1",
-    "react": "16.13.1",
-    "react-native": "0.63.1",
-    "react-native-permissions": "^2.1.4"
+    "react": "^17.0.1",
```
chrisfromwork, line 20: @ryantrem are there any limitations to what react/react-native version we can use? What is required for testing when moving to a newer version?

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineViewManager.cpp
```diff
@@ -0,0 +1,148 @@
...
+namespace winrt::BabylonReactNative::implementation {
+    EngineViewManager::EngineViewManager() {}
+
+    // IViewManager
+    hstring EngineViewManager::Name() noexcept {
+        return L"EngineView";
+    }
+
+    FrameworkElement EngineViewManager::CreateView() noexcept {
+        _swapChainPanel = SwapChainPanel();
```
ryantrem, line 25: I would sublcass `SwapChainPanel` and put most of this logic in that class, then have `EngineViewManager::CreateView` just be a super simple factory function. This is how it works on the other platforms. If you did that, then these revoker tokens would not be necessary as the event handler (the `SwapChainPanel` subclass) would have the same lifetime as the event source (the `SwapChainPanel` subclass). You'd still have to manage the `event_token` returned from `CompositionTarget::Rendering`, but you could do that in Loaded/Unloaded events of the `SwapChainPanel` sublcass.

Modules/@babylonjs/react-native/windows/BabylonReactNative/readme.txt
```diff
@@ -0,0 +1,23 @@
+========================================================================
```
ryantrem, line 1: Seems like this file could be removed? Is it really relevant to this project?

Modules/@babylonjs/react-native/windows/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
+# cmake 3.15.3+ because of a c++ regression when swift is enable
```
ryantrem, line 1: I'm confused about the relationship between this CMakeLists.txt and the checked in vcxproj that looks at least partially hand authored.

Modules/@babylonjs/react-native/windows/README.md
```diff
@@ -0,0 +1,7 @@
+# Build Steps:
```
ryantrem, line 1: I don't think we have platform specific readmes currently. Can this just be merged into the readme one level up?

ryantrem (commented):
Super awesome to see this progress!

I guess with a separate change we will tackle adding prebuilt binaries etc. to the NPM package?

Apps/Playground/windows/Playground/Package.appxmanifest
```diff
@@ -0,0 +1,53 @@
...
+        <uap:DefaultTile Wide310x150Logo="Assets\Wide310x150Logo.png"/>
+        <uap:SplashScreen Image="Assets\SplashScreen.png" />
+      </uap:VisualElements>
+    </Application>
+  </Applications>
+
+  <Capabilities>
+    <Capability Name="internetClient" />
+    <uap2:Capability Name="spatialPerception"/>
+    <DeviceCapability Name="webcam"/>
```
tymlipari, line 51: Do we actually use the webcam permission at all?

Modules/@babylonjs/react-native/windows/BabylonReactNative/BabylonModule.h
```diff
@@ -0,0 +1,21 @@
...
+    REACT_MODULE(BabylonModule, L"BabylonModule");
+    struct BabylonModule : std::enable_shared_from_this<BabylonModule>
+    {
+        REACT_INIT(Initialize);
+        void Initialize(const winrt::Microsoft::ReactNative::ReactContext& reactContext) noexcept;
+
+        REACT_METHOD(CustomInitialize, L"initialize");
+        void CustomInitialize(const winrt::Microsoft::ReactNative::ReactPromise<bool>& result) noexcept;
+
+        BabylonModule();
```
tymlipari, line 15: If these have empty implementations, can they be omitted?

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineViewManager.cpp
```diff
@@ -0,0 +1,148 @@
...
+        _swapChainPanelPtr = nullptr;
+        copy_to_abi(_swapChainPanel, _swapChainPanelPtr);
```
tymlipari, line 27: `copy_to_abi` ends up invoking `AddRef` on the underlying object. It looks like there's no corresponding place where we call `Release` meaning the swap chain panel is probably being leaked.

I'm not quite sure what you intend the ownership semantics of `_swapChainPanelPtr` to be. If you're only looking to get access to a `void*` to pass across an ABI boundary or something, but without changing the ref-count, I think it's fine to just use `get_abi`. If you're intending to increment the ref-count, make sure you're using something like `attach_abi` or something on the other side to ensure the lifetime is handled correctly.

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineViewManager.cpp
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    void EngineViewManager::OnSizeChanged(IInspectable const& /*sender*/, SizeChangedEventArgs const& args)
+    {
+        const auto size = args.NewSize();
+        _swapChainPanelWidth = static_cast<size_t>(size.Width);
+        _swapChainPanelHeight = static_cast<size_t>(size.Height);
+
+        // Use windowTypePtr == 2 for xaml swap chain panels
+        auto windowTypePtr = reinterpret_cast<void*>(2);
```
tymlipari, line 110: This is kind of unfortunate... We don't have a way to make this like an enum or something?

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineViewManager.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+        Babylon::UpdateView(_swapChainPanelPtr, _swapChainPanelWidth, _swapChainPanelHeight, windowTypePtr);
+    }
+
+    void EngineViewManager::OnPointerPressed(IInspectable const& /*sender*/, PointerRoutedEventArgs const& args)
+    {
+        const auto pointerId = args.Pointer().PointerId();
+        const auto buttonId = 0; // Update as needed
+        const auto point = args.GetCurrentPoint(_swapChainPanel);
+        const uint32_t x = point.Position().X < 0 ? 0 : point.Position().X;
```
tymlipari, line 120: nit: just because I'm not sure the compiler is smart enough to cache this after multiple calls since they're virtual at the ABI level, you should cache the result of `point.Position()` so you're not invoking the property getter multiple times.

tymlipari: Ditto for other events

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineViewManager.h
```diff
@@ -0,0 +1,73 @@
...
+
+    struct EngineViewManager
+        : winrt::implements<
+        EngineViewManager,
+        winrt::Microsoft::ReactNative::IViewManager,
+        winrt::Microsoft::ReactNative::IViewManagerWithReactContext,
+        winrt::Microsoft::ReactNative::IViewManagerWithNativeProperties,
+        winrt::Microsoft::ReactNative::IViewManagerWithExportedEventTypeConstants,
+        winrt::Microsoft::ReactNative::IViewManagerWithCommands>
+        , std::enable_shared_from_this<EngineViewManager> {
```
tymlipari, line 17: Don't combine `winrt::implements` with `std::enable_shared_from_this` since WinRT has its own reference counting mechanism that the STL's will conflict with. You should still be able to get weak pointers from C++/WinRT using `get_weak()`.

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineViewManager.h
```diff
@@ -0,0 +1,73 @@
...
+    public:
+        EngineViewManager();
+
+        // IViewManager
+        winrt::hstring Name() noexcept;
```
tymlipari, line 22: nit: for functions like `Name()` that are just getters, make them `const` wherever possible.

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineViewManager.h
```diff
@@ -0,0 +1,73 @@
...
+        size_t _swapChainPanelWidth{ 1 };
+        size_t _swapChainPanelHeight{ 1 };
+
+        struct RevokerData
+        {
+            winrt::Windows::UI::Xaml::FrameworkElement::SizeChanged_revoker SizeChangedRevoker{};
+            winrt::Windows::UI::Xaml::FrameworkElement::PointerPressed_revoker PointerPressedRevoker{};
+            winrt::Windows::UI::Xaml::FrameworkElement::PointerMoved_revoker PointerMovedRevoker{};
+            winrt::Windows::UI::Xaml::FrameworkElement::PointerReleased_revoker PointerReleasedRevoker{};
+            winrt::event_token RenderingToken;
```
tymlipari, line 67: You should be able to use `winrt::Windows::UI::Xaml::Media::CompositionTarget::Rendering_revoker` here even though it's a static event.

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineViewManager.cpp
ryantrem: We've talked about ways to make this type safe per platform in Babylon Native, but haven't fleshed out or implemented those ideas.

.vscode/settings.json
```diff
@@ -1,3 +1,6 @@
 {
-    "react-native-tools.projectRoot": "./Apps/Playground"
+    "react-native-tools.projectRoot": "./Apps/Playground",
+    "files.associations": {
+        "*.ipp": "cpp"
+    }
```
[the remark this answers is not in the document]
chrisfromwork: I don't believe this is needed. reverting 

---
In reply to: [557643597](https://github.com/BabylonJS/BabylonReactNative/pull/137#discussion_r557643597) [](ancestors = 557643597)

Apps/Playground/metro.config.js
```diff
@@ -62,14 +62,24 @@ module.exports = {
...
+      // Provide the set of known local package mappings.
+      moduleMappings,
+      {
+        // Provide a mapper function, which uses the above mappings for associated package ids,
+        // otherwise fall back to the standard behavior and just look in the node_modules directory.
+        get: (target, name) => name in target ? target[name] : path.join(__dirname, `node_modules/${name}`),
+      },
     ),
+    
+    blockList: exclusionList([
```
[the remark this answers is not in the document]
chrisfromwork: For some reason react-native-windows moved from blacklists to blocklists in December. I'll reach out to react-native-windows owners to see why

---
In reply to: [557654635](https://github.com/BabylonJS/BabylonReactNative/pull/137#discussion_r557654635) [](ancestors = 557654635)

Modules/@babylonjs/react-native/windows/BabylonReactNative/readme.txt
chrisfromwork: This file is auto-added when installing winrt c++. Will remove

Modules/@babylonjs/react-native/windows/CMakeLists.txt
chrisfromwork: From what i've heard, cmake can't generate a single visual studio solution for multiple platforms/configurations, so we have a hybrid solution:

1. cmake generates BabylonNative dependencies/static libs (for example: BabylonReactNative\Modules\@babylonjs\react-native\submodules\BabylonNative\Build_uwp_x86\ReactNativeBabylon.sln)
2. We manually reference all static libraries for all platforms in the react-native NativeModule project (BabylonReactNative\Modules\@babylonjs\react-native\windows\BabylonReactNative.sln). This allows us to build multiple platform/configurations from the same solution/vcxproj definition.

FWIW, react-native-windows NativeModules take some manual setup/manual edits to vcxproj definitions. We could theoretically do this through cmake, but this requirement in addition to cmake needing to then generate 4+ projects/solutions just to build all the platform flavors we need made it seem like manually setting up a BabylonReactNative.dll project definition was a better developer experience.

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineViewManager.cpp
chrisfromwork: FWIW, this code path doesn't seem well documented in bgfx. It already feels somewhat like a hack where folks overloaded a pointer needed for linux logic to do something random for windows logic.

Modules/@babylonjs/react-native/windows/README.md
chrisfromwork: yeah sounds good

Apps/Playground/metro.config.js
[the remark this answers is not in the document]
chrisfromwork: Talking to some different folks, it sounds like this was a rename in the metro version included with react-native 0.64. It should do the same things as before and is just a rename

---
In reply to: [560367678](https://github.com/BabylonJS/BabylonReactNative/pull/137#discussion_r560367678) [](ancestors = 560367678,557654635)

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineViewManager.cpp
```diff
@@ -0,0 +1,148 @@
...
+        return commands.GetView();
+    }
+
+    void EngineViewManager::DispatchCommand(
+        FrameworkElement const& view,
+        winrt::hstring const& commandId,
+        IJSValueReader const& /*commandArgsReader*/) noexcept {
+        if (auto swapChainPanel = view.try_as<SwapChainPanel>()) {
+            if (commandId == L"takeSnapshot") {
+                // TODO: Implement takeSnapshot
```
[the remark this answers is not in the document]
chrisfromwork: sounds good

Modules/@babylonjs/react-native/windows/BabylonReactNative/BabylonModule.h
chrisfromwork: yes they can

Apps/Playground/windows/Playground/Package.appxmanifest
chrisfromwork: no we don't need this

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineViewManager.h
chrisfromwork: looks like we were never using the standard library logic and this is a stale inheritance

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineViewManager.cpp
chrisfromwork: sounds good

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineViewManager.cpp
chrisfromwork: i think we should just use get_abi here and allow EngineViewManager to maintain SwapChainPanel lifecycle. thanks for the additional context!

Apps/Playground/windows/Playground/App.cpp
```diff
@@ -0,0 +1,95 @@
...
+    InstanceSettings().DebuggerBreakOnNextLine(true);
+    InstanceSettings().UseFastRefresh(false);
+#endif
+
+    RegisterAutolinkedNativeModulePackages(PackageProviders()); // Includes any autolinked modules
+
+    PackageProviders().Append(make<ReactPackageProvider>()); // Includes all modules in this project
+
+    // TODO: we should use autolinking once possible
+    PackageProviders().Append(winrt::SliderWindows::ReactPackageProvider());
```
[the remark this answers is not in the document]
chrisfromwork: This content should get automatically added to AutolinkedNativeModules.g.cpp. For some reason this isn't the case with the current community slider release. I've filed issues against the repo, but we need to do this for now.

If we don't want this code here we need to remove the slider from App.tsx, which I don't think is worth the hassle.

Modules/@babylonjs/react-native/package.json
```diff
@@ -28,17 +28,27 @@
   },
   "peerDependencies": {
     "@babylonjs/core": "^5.0.0-alpha.6",
-    "react": "^16.13.1",
-    "react-native": "^0.63.1",
-    "react-native-permissions": "^2.1.4"
+    "react": "^17.0.1",
+    "react-native": "^0.64.0-rc.0",
```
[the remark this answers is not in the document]
chrisfromwork: is there any mechanism for defining different dependencies based on platform?

Modules/@babylonjs/react-native/windows/BabylonReactNative/BabylonReactNative.vcxproj
```diff
@@ -0,0 +1,625 @@
+<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
chrisfromwork: yes. from what i've heard, cmake can't be used to generate projects to build for multiple platforms/configurations

Modules/@babylonjs/react-native/windows/BabylonReactNative.sln
```diff
@@ -0,0 +1,3613 @@
+﻿
```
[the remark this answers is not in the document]
chrisfromwork: Yes we need to keep this. This solution references all static lib project files for BabylonNative dependencies. Playground.sln does not. We could change this as needed, but I think that it would be better in a follow up review.

Modules/@babylonjs/react-native/windows/BabylonReactNative/BabylonModule.cpp
```diff
@@ -0,0 +1,40 @@
...
+                    trueThis->_reactContext.JSDispatcher().Post([weakThis, func{ std::move(func) }]() {
+                        func();
+                    });
+                }
+            };
+            Babylon::Initialize(jsiRuntime, jsDispatcher, false);
+        }
+    });
+
+    result.Resolve(true);
```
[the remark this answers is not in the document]
chrisfromwork: FWIW, CustomInitialize is called on the js thread and ExecuteJsi doesn't currently queue functions, but your suggestion is better if ExecuteJsi changes in the future.

Modules/@babylonjs/react-native/windows/BabylonReactNative/BabylonReactNative.vcxproj
```diff
@@ -0,0 +1,625 @@
...
+    <AppContainerApplication>true</AppContainerApplication>
+    <ApplicationType>Windows Store</ApplicationType>
+    <ApplicationTypeRevision>10.0</ApplicationTypeRevision>
+    <WindowsTargetPlatformVersion Condition=" '$(WindowsTargetPlatformVersion)' == '' ">10.0.18362.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformMinVersion>10.0.18362.0</WindowsTargetPlatformMinVersion>
+  </PropertyGroup>
+  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
+  <PropertyGroup Label="ReactNativeWindowsProps">
+    <!-- First try and resolve react-native-windows package at a relative location when installing the @babylonjs/react-native package in your project -->
+    <ReactNativeWindowsDir Condition="'$(ReactNativeWindowsDir)' == '' And Exists('$(ProjectDir)..\..\..\..\..\node_modules\react-native-windows\package.json')">$(ProjectDir)\..\..\..\..\..\node_modules\react-native-windows\</ReactNativeWindowsDir>
```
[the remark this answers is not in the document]
chrisfromwork: I think we should allow for both for now. I'm not sure if we want to bloat the Playground.sln with all static lib project references (x86, arm, arm64, etc) for BabylonNative dependencies. It turns into 140+ projects and takes a long time to load.

I'd be happy to relook this again when looking into packaging an npm package for UWP.

Modules/@babylonjs/react-native/windows/BabylonReactNative/BabylonModule.cpp
```diff
@@ -0,0 +1,40 @@
...
+            {
+                if (auto trueThis = weakThis.lock())
+                {
+                    trueThis->_reactContext.JSDispatcher().Post([weakThis, func{ std::move(func) }]() {
+                        func();
+                    });
+                }
+            };
+            Babylon::Initialize(jsiRuntime, jsDispatcher, false);
+        }
```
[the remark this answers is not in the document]
chrisfromwork: sounds good

Modules/@babylonjs/react-native/package.json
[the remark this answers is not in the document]
chrisfromwork: I will update the readme.txt to call out this discrepancy

Apps/Playground/windows/Playground/AutolinkedNativeModules.g.cpp
```diff
@@ -0,0 +1,23 @@
+// AutolinkedNativeModules.g.cpp contents generated by "react-native autolink-windows"
```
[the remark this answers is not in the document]
chrisfromwork: npm install doesn't run autolinking, but npm run windows does. Looking at react-native-windows samples this file is checked in. I would prefer to mimic that behavior.

Apps/Playground/App.tsx
```diff
@@ -91,6 +91,12 @@ const EngineScreen: FunctionComponent<ViewProps> = (props: ViewProps) => {
           const session = await xr.baseExperience.enterXRAsync("immersive-ar", "unbounded", xr.renderTarget);
           setXrSession(session);
 
+          if (Platform.OS == "windows") {
+            // HMDs require different rendering behaviors than default xr rendering for mobile devices
+            scene.clearColor = new Color4(0, 0, 0, 1); // Black
+            scene.autoClear = true;
+          }
```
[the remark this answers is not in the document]
chrisfromwork: this has now been moved to enginehook.ts and a follow up issues has been filed in the BabylonNative repo

Apps/Playground/package.json
```diff
@@ -13,29 +15,35 @@
...
+    "react-native-permissions": "^3.0.0",
+    "react-native-windows": "preview"
   },
   "devDependencies": {
     "@babel/core": "^7.8.4",
     "@babel/runtime": "^7.8.4",
     "@react-native-community/cli": "^4.13.0",
     "@react-native-community/eslint-config": "^1.1.0",
+    "@rnw-scripts/eslint-config": "0.1.6",
+    "@rnw-scripts/ts-config": "0.1.0",
```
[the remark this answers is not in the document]
chrisfromwork: these were defaults in react-native-windows samples but don't appear to be needed. they've been removed.

Apps/Playground/package.json
```diff
@@ -13,29 +15,35 @@
...
+    "eslint": "7.12.0",
     "jest": "^25.1.0",
     "metro-react-native-babel-preset": "^0.59.0",
     "prettier": "^2.0.4",
-    "react-test-renderer": "16.13.1",
-    "typescript": "^3.8.3"
+    "react-test-renderer": "^17.0.1",
+    "typescript": "^3.8.3",
+    "just-scripts": "^0.44.7",
+    "metro-config": "^0.64.0"
```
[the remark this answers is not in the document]
chrisfromwork: these were defaults in react-native-windows samples but don't appear to be needed. they've been removed.

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineViewManager.cpp
chrisfromwork: content has been moved to a new class

Modules/@babylonjs/react-native/EngineHook.ts
```diff
@@ -63,6 +63,17 @@ class DOMException {
     }
 }
 
+if (Platform.OS == "windows") {
+    const originalEnterXRAsync: (...args: any[]) => Promise<any> = WebXRExperienceHelper.prototype.enterXRAsync;
+    WebXRExperienceHelper.prototype.enterXRAsync = async function (...args: any[]): Promise<any> {
+        // TODO: https://github.com/BabylonJS/BabylonNative/issues/577
+        // Windows HMDs require different rendering behaviors than default xr rendering for mobile devices
+        await originalEnterXRAsync.apply(this, args);
+        this.scene.clearColor = new Color4(0, 0, 0, 1); // Black
```
ryantrem, line 72:
```suggestion
        this.scene.clearColor = Color3.Black().toColor4();
```

Modules/@babylonjs/react-native/windows/BabylonReactNative/BabylonModule.cpp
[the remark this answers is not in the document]
ryantrem: Seems like you didn't actually change it though?

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineViewManager.h
```diff
@@ -38,35 +35,9 @@ namespace winrt::BabylonReactNative::implementation {
...
-        struct RevokerData
-        {
-            winrt::Windows::UI::Xaml::FrameworkElement::SizeChanged_revoker SizeChangedRevoker{};
-            winrt::Windows::UI::Xaml::FrameworkElement::PointerPressed_revoker PointerPressedRevoker{};
-            winrt::Windows::UI::Xaml::FrameworkElement::PointerMoved_revoker PointerMovedRevoker{};
-            winrt::Windows::UI::Xaml::FrameworkElement::PointerReleased_revoker PointerReleasedRevoker{};
-            winrt::event_token RenderingToken;
-        };
-        RevokerData _revokerData{};
+        winrt::BabylonReactNative::EngineView _engineView{ nullptr };
```
ryantrem, line 40: The only reason this instance is stored is so that it can be replaced, which destroys the old one and unregisters events, right? If/when in the future we add support for multiple view instances, we'll have to update this, and probably do something more like the `OnLoaded`/`OnUnloaded` approach I mentioned, but I guess we can just do it this way for now and worry about fixing it up as needed in the future.

Package/gulpfile.js
```diff
@@ -40,6 +40,16 @@ const buildAndroid = async () => {
...
 
+const makeUWPProject = async () => {
+  exec('.\\..\\Modules\\@babylonjs\\react-native\\windows\\scripts\\Setup.bat');
+}
+
+const buildUWPProject = async () => {
+  exec('.\\..\\Modules\\@babylonjs\\react-native\\windows\\scripts\\Build.bat');
+}
+
+const buildUWP = gulp.series(makeUWPProject, buildUWPProject);
```
ryantrem, line 51: Today we run this gulp build on Mac agents so we can build both iOS and Android, which means this added stuff for UWP will break. I think we are going to need to rework this so that we build Android, iOS, and UWP on different build agents, publish the built stuff as artifacts, and then have a separate stage that pulls the artifacts and packages them all together into the NPM package. Maybe we just save the changes to gulpfile.js for a future PR?

Modules/@babylonjs/react-native/windows/CMakeLists.txt
ryantrem: Ok, so the process right now is to manually run a script or something that runs CMake to generate solutions/projects for each platform/configuration, then build all those, then build the hand crafted solution that references the outputs (built libs) of those generated solutions/projects?

Package/gulpfile.js
chrisfromwork: that sounds good. I'll probably go ahead and add the buildUWP gulp task to the exports. But i agree it shouldn't be in the main build task.

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineViewManager.h
chrisfromwork: yeah we can refactor this as wanted. right now it seems like BabylonNative initialization is driven by the BabylonModule initialie call. I wasn't sure if we wanted the BabylonNative initialization/deinitialization to be driven by the SwapChainPanel OnLoaded/OnUnloaded.

FWIW, there are some lifecycle bugs floating around for UWP support. We can address these sorts of changes in these follow up fixes.

Modules/@babylonjs/react-native/windows/CMakeLists.txt
chrisfromwork: yeah thats the general workflow. If you open the BabylonReactNative.sln in Modules\@babylonjs\react-native you can build everything. Playground.sln doesn't currently reference the static lib vcxprojs, so it can only build the BabylonReactNative dll and assumes all libs are already compiled/populated. For packaging, we will ship the source/vcxproj for BabylonReactNative.dll + precompiled libs. I am still tinkering with the vcxproj setup to support this.

Apps/Playground/App.tsx
```diff
@@ -6,7 +6,7 @@
  */
 
 import React, { useState, FunctionComponent, useEffect, useCallback } from 'react';
-import { SafeAreaView, StatusBar, Button, View, Text, ViewProps, Image } from 'react-native';
+import { SafeAreaView, StatusBar, Button, View, Text, ViewProps, Image, Platform } from 'react-native';
```
ryantrem, line 9: I don't think this file is using `Platform` any more.
```suggestion
import { SafeAreaView, StatusBar, Button, View, Text, ViewProps, Image } from 'react-native';
```

ryantrem (approved):

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineView.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+using namespace winrt::Windows::UI::Xaml;
+using namespace winrt::Windows::UI::Xaml::Input;
+using namespace winrt::Windows::UI::Xaml::Media;
+using namespace winrt::Windows::UI::Xaml::Controls;
+
+namespace winrt::BabylonReactNative::implementation {
+    EngineView::EngineView() {
+        _revokerData.SizeChangedRevoker = SizeChanged(winrt::auto_revoke, { this, &EngineView::OnSizeChanged });
+        _revokerData.PointerPressedRevoker = PointerPressed(winrt::auto_revoke, { this, &EngineView::OnPointerPressed });
```
tymlipari, line 13: For pointer events, rather than subscribing to the evented form, you should instead override the `OnPointer`* methods. These are invoked before the evented forms, and provide an option for suppressing the event from being raised (by setting the `Handled` property) in cases where that might not be valuable (think not raising `PointerPressed` for a button but instead a `Click`).

tymlipari (approved):

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineView.cpp
chrisfromwork: sounds good, I'll add this to the upcoming packaging review.

PR #138, opened 2021-01-12
Min support for manual rendering (as an implementation detail)

ryantrem:
This change adds the minimum needed to easily support manual rendering in the shared Babylon Native / React Native integration layer. Android and iOS continue to use automatic rendering, but this should let us use the shared code for the React Native for Windows integration (where we need to do manual rendering). Later we should expose something like `BeginRenderingView`/`EndRenderingView` to allow for more efficient render loops, but the RNW integration as is does not use that, so that's an improvement we can make later. We also need to sort out how to do the `ResetView` logic in the manual rendering case, but that already has other issues we need to figure out in the RNW integration. So again, min changes needed to bring the in-progress RNW integration onto the shared code.

### DIFF
--- a/Modules/@babylonjs/react-native/shared/BabylonNative.cpp
+++ b/Modules/@babylonjs/react-native/shared/BabylonNative.cpp
@@ -14,12 +14,18 @@ namespace Babylon
 {
     using namespace facebook;
 
+    namespace
+    {
+        Dispatcher g_inlineDispatcher{[](const std::function<void()>& func) { func(); }};
+    }
+
     class ReactNativeModule : public jsi::HostObject
     {
     public:
-        ReactNativeModule(jsi::Runtime& jsiRuntime, Dispatcher jsDispatcher)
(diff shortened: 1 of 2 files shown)


Modules/@babylonjs/react-native/shared/BabylonNative.cpp
```diff
@@ -14,12 +14,18 @@ namespace Babylon
 {
     using namespace facebook;
 
+    namespace
+    {
+        Dispatcher g_inlineDispatcher{[](const std::function<void()>& func) { func(); }};
```
chrisfromwork, line 19: >g_inlineDispatcher [](start = 19, length = 18)

is the thought here to have synchronous execution on the calling thread?

chrisfromwork (approved):
This seems like it should address UWP needs

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
ryantrem: Yes that's exactly right. It just makes it a little easier to write the conditional logic in `UpdateView`.

PR #142, opened 2021-01-22
Add uwp packaging logic

chrisfromwork:
This review adds uwp packaging logic to the pre-existing android/ios NPM package. It contains the following:

1. Modules\@babylonjs\react-native\windows\BabylonReactNative.sln has been removed. We will plan to build static libraries for the cmake generated BabylonNative solution (ReactBabylonNative.sln). We will build the BabylonReactNative.dll when it is autolinked into a react-native-windows project. Trying to prebuild the BabylonReactNative.dll seems like it will be prone to issues if folks have mismatching versions of react-native and react-native-windows.
1. Build.ps1 has been updated to build the cmake generated solution to generate static libs compared to building BabylonReactNative.sln
1. gulp tasks have been created to build and pack uwp resources. These tasks are defined independent to any preexisting android/iOS packaging tasks
1. PackageTest has been updated to support react-native-windows. It now also consumes generated uwp NPM packages.
1. The metro-config NPM package has been added back as a dev dependency for the Playground app. It seems that without this package declaration, metro is unable to resolve exclusionList. I believe in the past I may have had stale NPM packages allowing this to still work without declaring the devDependency


Apps/Playground/package.json
```diff
@@ -18,7 +18,7 @@
     "@react-native-community/slider": "4.0.0-rc.2",
     "logkitty": "^0.7.1",
     "react": "^17.0.1",
-    "react-native": "^0.64.0-rc.0",
+    "react-native": "0.64.0-rc.0",
```
ryantrem, line 21: Why is this change needed?

Package/package.json
```diff
@@ -3,6 +3,10 @@
   "devDependencies": {
     "fancy-log": "^1.3.3",
     "gulp": "^4.0.2",
+    "gulp-rename": "2.0.0",
     "shelljs": "^0.8.4"
+  },
+  "dependencies": {
+    "@babylonjs/react-native": "file:Assembled/babylonjs-react-native-0.0.1.tgz"
```
ryantrem, line 10: Guessing this was inadvertent... revert?

ryantrem:
Regarding this comment in your PR description:

> We will build the BabylonReactNative.dll when it is autolinked into a react-native-windows project.

Autolinking makes the `BabylonReactNative.vcxproj` get referenced as a dependency by the consuming RNW app, and the associated `BabylonReactNative.dll` is built when the consuming RNW app is built, correct?

Package/gulpfile.js
```diff
@@ -93,6 +94,89 @@ const copyAndroidFiles = async () => {
...
+    .pipe(rename({ dirname: '' }))
+    .pipe(gulp.dest('Assembled/windows/libs/arm64/Release'));
+}
+
+const copyVCXProjUWPFiles = () => {
+  return gulp.src('../Modules/@babylonjs/react-native/windows/BabylonReactNative/*.*')
+    .pipe(gulp.dest('Assembled/windows/BabylonReactNative'));
+}
+
+const copyUWPFiles = gulp.series(
```
ryantrem, line 168: Could this one actually be `gulp.parallel`?

Apps/PackageTest/metro.config.js
```diff
@@ -4,13 +4,25 @@
...
+  resolver: {
+    blockList: exclusionList([
+      // This stops "react-native run-windows" from causing the metro server to crash if its already running
+      new RegExp(
+        `${path.resolve(__dirname, 'windows').replace(/[/\\]/g, '/')}.*`,
+      ),
+      // This prevents "react-native run-windows" from hitting: EBUSY: resource busy or locked, open msbuild.ProjectImports.zip
+      /.*\.ProjectImports\.zip/,
+    ]),
+  },
```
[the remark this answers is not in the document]
chrisfromwork: react-native-windows on setup overwrites this file to contain this logic, and when other custom things are in this file it doesn't seem to merge them. If there are thoughts on how things could be improved, we should relay this feedback to the react-native-windows team.

Apps/Playground/package.json
chrisfromwork: when running react-native-windows initialization, react-native-windows will prompt/force users to use newer preview releases that support whatever react-native release is in the project. I don't think we are guaranteed that react-native-windows preview releases won't contain breaking changes (I have seen breaking changes already).

What I was seeing was that react-native would get installed at 0.64.0-rc.2, forcing us to update to react-native-windows 0.64.0-preview.7 which contained breaking changes for native modules (This was all during initialization prompts compared to an action of npm package installs). I believe we can ensure things will compile if we don't auto update to newer versions of react-native/react-native-windows. I think this will probably get better when react-native-windows has a formal 0.64.0 release.

Package/gulpfile.js
chrisfromwork: yeah seems like it.

Apps/PackageTest/metro.config.js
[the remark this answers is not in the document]
ryantrem: Yea that's pretty bad. I guess their assumption is that people will diff the file and manually fix it up. One thing that would help would be to put this logic in a separate file that is bundled with RNW and import it into this file. That might be good feedback for the rnw folks.

ryantrem (approved):

PR #144, opened 2021-01-25
Add windows support to github actions (new branch)

chrisfromwork:
1. PR validation is updated to build a x64 debug flavor of UWP support
2. Package generation now builds and includes UWP BabylonNative static libs and the BabylonReactNative solution
3. Lookup logic for msbuild now looks for the command in the developer's path variable. This is needed to resolve msbuild in github actions


.github/workflows/pr.yml
```diff
@@ -40,3 +40,26 @@ jobs:
...
         run: npx gulp buildIOS
         working-directory: ./Package
+
+  build-windows:
+    runs-on: windows-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2.3.3
+        with:
+          submodules: 'true'
```
ryantrem, line 50: Is this `true` instead of `recursive` specifically for the ARCore dependency? Don't we have other nested submodules, like bgfx under bgfx.cmake? How does this work?

.github/workflows/pr.yml
```diff
@@ -40,3 +40,26 @@ jobs:
...
+      - name: Add msbuild to PATH
```
ryantrem, line 51: I would probably just name this "Setup MSBuild" to match the other naming.

.github/workflows/pr.yml
```diff
@@ -40,3 +40,26 @@ jobs:
...
+        uses: microsoft/setup-msbuild@v1.0.2
+      - name: Add nuget to PATH
```
ryantrem, line 53: Maybe same thing here, "Setup NuGet"

.github/workflows/publish.yml
```diff
@@ -22,6 +22,54 @@ jobs:
...
+        with:
+          name: 'Assembled'
+          path: Package/Assembled
+  build-windows:
+    runs-on: windows-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2.3.3
+        with:
+          submodules: 'true'
```
ryantrem, line 36: Same question as above.

.github/workflows/publish.yml
```diff
@@ -22,6 +22,54 @@ jobs:
...
+      - name: Add msbuild to PATH
+        uses: microsoft/setup-msbuild@v1.0.2
+      - name: Add nuget to PATH
+        uses: nuget/setup-nuget@v1
+        with:
+          nuget-version: '5.x'
```
ryantrem, line 42: Same comments on naming as above, though I don't feel super strongly about it, so up to you.

.github/workflows/publish.yml
```diff
@@ -22,6 +22,54 @@ jobs:
...
+      - name: NPM Install (Binary Package)
+        run: npm install
+        working-directory: ./Package
+      - name: Download Assembled Folder
+        uses: actions/download-artifact@v2
+        with:
+          name: 'Assembled'
+          path: Package/Assembled
+      - name: Display structure of downloaded Assembled folder
+        run: ls -R
```
ryantrem, line 72: Is this just for debugging?

Modules/@babylonjs/react-native/windows/scripts/PRSetup.ps1
```diff
@@ -0,0 +1,16 @@
+Import-Module $PSScriptRoot\Utils.psm1
+
+# windows build agents don't support the path lengths required for initializing arcore dependencies,
+# so we manually initialize the submodules we need here.
```
ryantrem, line 4: Ok, I think this answers my earlier question about the submodule update. Instead of having an inclusion list like this, could we have an exclusion list? That might make it easier to maintain. Seems like this is possible? https://stackoverflow.com/a/52185169/7318670

Modules/@babylonjs/react-native/windows/scripts/Setup.ps1
```diff
@@ -1,5 +1,18 @@
 Import-Module $PSScriptRoot\Utils.psm1
```
ryantrem, line 1: What is the advantage of having these separate PowerShell scripts as opposed to just having all this logic in `gulpfile.js`?

Modules/@babylonjs/react-native/windows/scripts/Utils.psm1
```diff
@@ -63,7 +69,7 @@ function Compile-Solution {
     )
 
     $MSBuild = Get-MSBuildPath
-    & "$MSBuild" /p:Configuration="$Configuration" /p:Platform="$Platform" $Solution
+    & "$MSBuild" /p:Configuration="$Configuration" /p:Platform="$Platform" /m:32 $Solution
```
ryantrem, line 72: Should this just be `/m`? My understanding is that that tells MSBuild to build in parallel, but lets it decide how much parallelism based on the number of cpu cores in the machine where it is actually building.

ryantrem: Also, within a single project, the vc++ compiler can compile c++ files in parallel with the `/MP` compiler switch. Can/do/should we use that as well?

.github/workflows/publish.yml
chrisfromwork: yeah this was just for debugging. without tokens, I haven't been able to confirm this completely works. This did reveal that all the content was copied.

.github/workflows/pr.yml
chrisfromwork: this should be handled by the powershell script as you noted later in the review

.github/workflows/publish.yml
chrisfromwork: see previous answer

Modules/@babylonjs/react-native/windows/scripts/PRSetup.ps1
chrisfromwork: updated to match suggestion

Modules/@babylonjs/react-native/windows/scripts/Setup.ps1
chrisfromwork: The only thing here currently that benefits from powershell scripts is the logic for resolving msbuild. For local developer setups where msbuild isn't added to the path variable, we download the vs installer powershell module which helps us resolve the directory to look in for msbuild. I'm not sure if something similar exists with gulp.

Modules/@babylonjs/react-native/windows/scripts/Utils.psm1
chrisfromwork: specifying processors was more confusion on my end. If msbuild can auto determine the correct number of processors, I would prefer that as well.

/MP is already set through cmake for majority if not all BabylonNative static libs. It is also set for BabylonReactNative.dll

chrisfromwork: removed the number of processors from m

Modules/@babylonjs/react-native/windows/scripts/Setup.ps1
chrisfromwork: I am going to keep some of the helper logic in the Utils.psm1 script, but i have reduced powershell usage. Setup is now conducted through gulp tasks. Only build logic is facilitated through powershell now.

Modules/@babylonjs/react-native/windows/scripts/Utils.psm1
tymlipari: FYI, you might want to consider enabling parallelism on the project level too. Looks like there's been some improvements added in Visual Studio 2019 16.3 - https://devblogs.microsoft.com/cppblog/improved-parallelism-in-msbuild/

ryantrem (approved):

Modules/@babylonjs/react-native/windows/scripts/Utils.psm1
chrisfromwork: Parallelism feels beyond the scope of this review. I've filed an issue here to track: https://github.com/BabylonJS/BabylonNative/issues/584

PR #145, opened 2021-01-25
Fix native input usage for uwp

chrisfromwork:
1. We move input handlers off the ui thread
2. We consume a newer NativeInput component in BabylonNative

We should not check in this review until the following is completed:

- [x] https://github.com/BabylonJS/BabylonNative/pull/583

we will also need to update the submodule commit to this BabylonNative pr commit once its completed.


Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineView.cpp
```diff
@@ -2,17 +2,38 @@
...
-        _revokerData.PointerReleasedRevoker = PointerReleased(winrt::auto_revoke, { this, &EngineView::OnPointerReleased });
+
+        WorkItemHandler workItemHandler([weakThis{ this->get_weak() }](IAsyncAction const& /* action */)
+        {
+            if (auto trueThis = weakThis.get())
+            {
+                auto deviceTypes = static_cast<CoreInputDeviceTypes>(
+                    static_cast<uint32_t>(Windows::UI::Core::CoreInputDeviceTypes::Mouse) |
+                    static_cast<uint32_t>(Windows::UI::Core::CoreInputDeviceTypes::Touch) |
+                    static_cast<uint32_t>(Windows::UI::Core::CoreInputDeviceTypes::Pen));
```
ryantrem, line 25: The Babylon.js `DeviceSourceManager` distinguishes between these (at least touch and mouse - see https://github.com/BabylonJS/Babylon.js/blob/master/src/DeviceInput/InputDevices/deviceTypes.ts). `NativeInput` is currently hard coded to always raise events for `Touch` only (that's just as much as has been implemented). I think what would probably make more sense would be to restrict this code to handle `Touch` only, or do more plumbing in `NativeInput` to support additional device types. If you want to try doing the latter, let me know and I'm happy to discuss.

chrisfromwork: I am pretty sure if we only support touch none of this logic will work for x64 desktop scenarios with a mouse. whats the suggestion for pen? having it register as touch?

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineView.cpp
```diff
@@ -2,17 +2,38 @@
...
+
+                trueThis->_revokerData.PointerPressedRevoker = coreInput.PointerPressed(winrt::auto_revoke, { trueThis.get(), &EngineView::OnPointerPressed });
+                trueThis->_revokerData.PointerMovedRevoker = coreInput.PointerMoved(winrt::auto_revoke, { trueThis.get(), &EngineView::OnPointerMoved });
+                trueThis->_revokerData.PointerReleasedRevoker = coreInput.PointerReleased(winrt::auto_revoke, { trueThis.get(), &EngineView::OnPointerReleased });
+
+                coreInput.Dispatcher().ProcessEvents(Windows::UI::Core::CoreProcessEventsOption::ProcessUntilQuit);
+            }
+        });
+
+        _inputLoopWorker = ThreadPool::RunAsync(workItemHandler, WorkItemPriority::High, WorkItemOptions::TimeSliced);
```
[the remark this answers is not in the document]
chrisfromwork: This was a suggestion from @tymlipari they may have more context

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineView.cpp
chrisfromwork: FWIW, input does not work currently. This change fixes that bug. It sounds like updating NativeInput involves feature work compared to bug fixes.

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineView.cpp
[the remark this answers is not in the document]
tymlipari: I don't think we have to right now, but `CoreIndependentInputSource` supports this in general. If we ever want to move the Babylon rendering off-thread, we can similarly move the input sink with it.

tymlipari:
Does pointer capture work here? Meaning, if I press down within the window and then drag the cursor outside the window bounds without releasing, does Babylon continue to receive the input? For things like camera manipulation and such that'd be a necessary feature, and we might need a mechanism to mark the pointers as captured until we're done with them.

ryantrem:
> Does pointer capture work here? Meaning, if I press down within the window and then drag the cursor outside the window bounds without releasing, does Babylon continue to receive the input? For things like camera manipulation and such that'd be a necessary feature, and we might need a mechanism to mark the pointers as captured until we're done with them.

Do we know how it works in the browser, and should we match whatever that behavior is?

Apps/Playground/App.tsx
```diff
@@ -50,6 +50,15 @@ const EngineScreen: FunctionComponent<ViewProps> = (props: ViewProps) => {
               }
             }
           })
+        } else if (device.deviceType === DeviceType.Mouse) {
```
ryantrem, line 53: Based on https://github.com/BabylonJS/BabylonNative/pull/583#discussion_r564143476, what this code should do is check for a pre-existing mouse device and if it exists, register for `onInputChangedObservable` on that device. When we see the mouse move, just poll the state of the left mouse button and see if it is down and only if so do the object rotation. This is how the code would be written when running in the browser. (Also it would be good to factor out the actual `rootNode.rotate(...)` into a local const lambda that can be called in both the touch and mouse cases so it's not repeated).

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineView.cpp
ryantrem: Yes, no work had previously been done to support mouse in `NativeInput`, so there is "feature" level work needed to support it correctly.

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
```diff
@@ -33,13 +34,13 @@ namespace Babylon
...
-                    m_resolveInitPromise = [&rt, resolve{ std::make_shared<jsi::Value>(rt, args[0]) }]()
-                    {
-                        resolve->asObject(rt).asFunction(rt).call(rt);
-                    };
-                    return {};
-                })
+                    resolve->asObject(rt).asFunction(rt).call(rt);
+                };
+                return {};
+            })
```
ryantrem, line 43: Revert? Seems like this got un-indented or something?

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineView.cpp
```diff
@@ -2,17 +2,47 @@
...
 using namespace winrt::Windows::UI::Xaml::Input;
 using namespace winrt::Windows::UI::Xaml::Media;
 using namespace winrt::Windows::UI::Xaml::Controls;
 
 namespace winrt::BabylonReactNative::implementation {
+    namespace {
+        constexpr uint32_t LEFT_MOUSE_BUTTON_ID = 0;
+        constexpr uint32_t MIDDLE_MOUSE_BUTTON_ID = 1;
+        constexpr uint32_t RIGHT_MOUSE_BUTTON_ID = 2;
+        constexpr uint32_t TOUCH_BUTTON_ID = 0;
```
ryantrem, line 20: These must match Babylon.js (which they do), and they must be the same on all platforms, so it seems like it would make more sense for these to be defined (and publicly exposed) in `NativeInput.h`. Otherwise we have to keep them in sync across platforms, right?

Apps/Playground/App.tsx
```diff
@@ -50,6 +50,15 @@ const EngineScreen: FunctionComponent<ViewProps> = (props: ViewProps) => {
...
+          const mouse: DeviceSource<DeviceType.Mouse> = deviceSourceManager.getDeviceSource(device.deviceType, device.deviceSlot)!;
+          mouse.onInputChangedObservable.add(mouseEvent => {
+            if (mouse.getInput(PointerInput.LeftClick) &&
+              mouseEvent.inputIndex === PointerInput.Horizontal &&
+              mouseEvent.currentState && mouseEvent.previousState) {
+              rootNode.rotate(Vector3.Down(), (mouseEvent.currentState - mouseEvent.previousState) * 0.005);
+            }
+          });
```
ryantrem, line 61: This is the code I would expect to work correctly (barring any typos):
```ts
      const deviceSourceManager = new DeviceSourceManager();

      const handlePointerInput = (pointer: DeviceSource<DeviceType.Mouse | DeviceType.Touch>) => {
        pointer.onInputChangedObservable.add(pointerEvent => {
          if (pointerEvent.inputIndex === PointerInput.Horizontal) {
            if (pointerEvent.currentState && pointerEvent.previousState) {
              rootNode.rotate(Vector3.Down(), (pointerEvent.currentState - pointerEvent.previousState) * 0.005);
            }
          }
        });
      };

      deviceSourceManager.onDeviceConnectedObservable.add(device => {
        if (device.deviceType === DeviceType.Touch) {
          const touch: DeviceSource<DeviceType.Touch> = deviceSourceManager.getDeviceSource(device.deviceType, device.deviceSlot)!;
          handlePointerInput(touch);
        }
      });

      const mouse = deviceSourceManager.getDeviceSource(DeviceType.Mouse);
      if (mouse) {
        handlePointerInput(mouse);
      }
```

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineView.cpp
chrisfromwork: EngineView doesn't directly reference NativeInput.h currently, so we may need a way to expose these values through BabylonNative.h. Regardless, it sounds good to have these values defined in NativeInput.h somehow

chrisfromwork: i'll take another look at how to propagate this

Apps/Playground/App.tsx
chrisfromwork: We will need some sort of hybrid solution here. I still see device connected events fired after this subscription has happened. This provided logic also doesn't assess pointer down states for mouse.

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineView.cpp
chrisfromwork: this should now be fixed

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
chrisfromwork: this should now be fixed

Apps/Playground/App.tsx
chrisfromwork: I think we need consensus on which of the following is the correct event behavior. Currently in the BabylonNative pr we do the first:

1. DeviceSourceManager is constructed, mouse is added
1. Something subscribes to device source manager's on connected event
1. DeviceSourceManager fires device connected callback for the mouse specifically for the new subscriber

vs. 

1. DeviceSourceManager is constructed, mouse is added
1. Something subscribes to the device source manager's on connected event
1. DeviceSourceManager does not fire device connected callback for the previously connected mouse. Subscriber should look at existing input devices via the device source manager to know the state of things before subscribing

the existing code will work with the first event callback behavior. I believe the code you have provided here is intended to support the second implementation choice.

Apps/Playground/App.tsx
```diff
@@ -50,6 +50,15 @@ const EngineScreen: FunctionComponent<ViewProps> = (props: ViewProps) => {
               }
             }
```
ryantrem, line 51: Will you just factor out this part to a local const lambda?
```ts
const handlePointerInput = (inputIndex: PointerInput, previousState: Nullable<number>, currentState: Nullable<number>) => {
  if (inputIndex === PointerInput.Horizontal &&
      currentState && previousState) {
        rootNode.rotate(Vector3.Down(), (currentState - previousState) * 0.005);
      };
};
```

Apps/Playground/App.tsx
ryantrem: Ok I didn't realize that `DeviceSourceManager` automatically invoked new callbacks of `onDeviceConnectedObservable` for existing devices. Given this, I think just the small refactor I mentioned above is good.

Modules/@babylonjs/react-native/android/src/main/cpp/BabylonNativeInterop.cpp
```diff
@@ -79,10 +79,10 @@ extern "C" JNIEXPORT void JNICALL Java_com_babylonreactnative_BabylonNativeInter
 
 extern "C" JNIEXPORT void JNICALL Java_com_babylonreactnative_BabylonNativeInterop_00024BabylonNative_setPointerButtonState(JNIEnv* env, jclass obj, jint pointerId, jint buttonId, jboolean isDown, jint x, jint y)
```
ryantrem, line 80: `buttonId` is no longer used, will you remove this?

Modules/@babylonjs/react-native/ios/BabylonNativeInterop.mm
```diff
@@ -79,21 +79,21 @@ + (void)reportTouchEvent:(MTKView*)mtkView touches:(NSSet<UITouch*>*)touches eve
                     } else {
                         [activeTouches replaceObjectAtIndex:pointerId withObject:touch];
                     }
-                    Babylon::SetPointerButtonState(static_cast<uint32_t>(pointerId), 0, true, x, y);
+                    Babylon::SetTouchButtonState(static_cast<uint32_t>(pointerId), 0, true, x, y);
```
ryantrem, line 82: Presumably this is what is causing your build failure:
```suggestion
                    Babylon::SetTouchButtonState(static_cast<uint32_t>(pointerId), true, x, y);
```

Modules/@babylonjs/react-native/ios/BabylonNativeInterop.mm
```diff
@@ -79,21 +79,21 @@ + (void)reportTouchEvent:(MTKView*)mtkView touches:(NSSet<UITouch*>*)touches eve
...
+                    Babylon::SetTouchPosition(static_cast<uint32_t>(pointerId), x, y);
                     break;
                 }
 
                 case UITouchPhaseEnded:
                 case UITouchPhaseCancelled: {
                     NSUInteger pointerId = [activeTouches indexOfObject:touch];
                     [activeTouches replaceObjectAtIndex:pointerId withObject:[NSNull null]];
-                    Babylon::SetPointerButtonState(static_cast<uint32_t>(pointerId), 0, false, x, y);
+                    Babylon::SetTouchButtonState(static_cast<uint32_t>(pointerId), 0, false, x, y);
```
ryantrem, line 96:
```suggestion
                    Babylon::SetTouchButtonState(static_cast<uint32_t>(pointerId), false, x, y);
```

Modules/@babylonjs/react-native/shared/BabylonNative.h
```diff
@@ -10,6 +10,12 @@ namespace Babylon
...
-    void SetPointerButtonState(uint32_t pointerId, uint32_t buttonId, bool isDown, uint32_t x, uint32_t y);
-    void SetPointerPosition(uint32_t pointerId, uint32_t x, uint32_t y);
+    void SetMouseButtonState(uint32_t buttonId, bool isDown, uint32_t x, uint32_t y);
+    void SetMousePosition(uint32_t x, uint32_t y);
+    void SetTouchButtonState(uint32_t pointerId, bool isDown, uint32_t x, uint32_t y);
+    void SetTouchPosition(uint32_t pointerId, uint32_t x, uint32_t y);
+
+    extern const uint32_t LEFT_MOUSE_BUTTON_ID;
+    extern const uint32_t MIDDLE_MOUSE_BUTTON_ID;
+    extern const uint32_t RIGHT_MOUSE_BUTTON_ID;
```
ryantrem, line 20: Just make these `constexpr`?

chrisfromwork: if we want these values to map to the values in NativeInput.h they need to be defined in a cpp file. Otherwise we have to reference the NativeInput header forcing all BabylonNative.h consumers to start having to reference NativeInput include folders.

if we just want to re-hardcode these values here we could use a constexpr, but that doesn't ensure we don't get out of sync with native input definitions

Apps/Playground/App.tsx
chrisfromwork: sounds good

Apps/Playground/App.tsx
chrisfromwork: this should now be fixed in the component. i'll follow the suggested refactor as well.

Modules/@babylonjs/react-native/shared/BabylonNative.h
ryantrem: Ah right, I think you said that elsewhere and I don't think it quite connected. I understand now and I think what you have makes sense.

Apps/Playground/App.tsx
chrisfromwork: sounds good

Modules/@babylonjs/react-native/android/src/main/cpp/BabylonNativeInterop.cpp
chrisfromwork: yes

ryantrem (approved):

PR #147, opened 2021-01-29
Fix android and ios support

chrisfromwork:
Moving to react native 0.64.0 preview releases has caused regressions in android/ios. This review fixes those issues.


Apps/Playground/ios/Podfile
```diff
@@ -4,24 +4,20 @@ require_relative '../node_modules/@react-native-community/cli-platform-ios/nativ
 platform :ios, '12.0'
 
 target 'Playground' do
-  permissions_path = '../node_modules/react-native-permissions/ios'
-  pod 'Permission-Camera', :path => "#{permissions_path}/Camera.podspec"
```
ryantrem, line 8: We need these lines (except with react-native-permissions 3.0 we need to remove the `.podspec`).

Apps/Playground/ios/Podfile
```diff
@@ -4,24 +4,20 @@ require_relative '../node_modules/@react-native-community/cli-platform-ios/nativ
...
-
-  use_react_native!(:path => config["reactNativePath"])
-
-  target 'PlaygroundTests' do
+  use_react_native!(
+    :path => config[:reactNativePath],
+    # to enable hermes on iOS, change `false` to `true` and then install pods
+    :hermes_enabled => false
+  )
+  target 'RnDiffAppTests' do
```
ryantrem, line 13: "RnDiffApp" is the name of the test app they use for the upgrade helper. Anywhere you see this needs to be replaced with "Playground".

Modules/@babylonjs/react-native/android/build.gradle
```diff
@@ -74,6 +74,20 @@ android {
         ndk {
             abiFilters "arm64-v8a", "armeabi-v7a", "x86"
         }
+        packagingOptions {
+            // TODO if def based on gradle version where we only do this for 4.0.0+?
```
chrisfromwork, line 78: we can probably remove this comment, libBabylonNative.so is the only lib we distribute and this exclude isn't included in the npm package build.gradle file. @ryantrem are we okay with having everyone build the libBabylonNative.so with gradle 4.1.0?

Package/Android/build.gradle
```diff
@@ -33,6 +33,7 @@ buildscript {
             jcenter()
         }
         dependencies {
+// TODO move to 4.1.0?
```
chrisfromwork, line 36: For me locally, updating this gradle version didn't seem needed. @ryantrem would you like to keep it at 3.4.1 or move it to 4.1.0? I personally am not sure what impacts the change has on npm package consumers

Package/gulpfile.js
```diff
@@ -112,7 +112,7 @@ const copyAndroidFiles = async () => {
   });
 
   await new Promise(resolve => {
-          gulp.src('../Apps/Playground/node_modules/@babylonjs/react-native/android/build/intermediates/library_and_local_jars_jni/release/**/*')
+          gulp.src('../Apps/Playground/node_modules/@babylonjs/react-native/android/build/intermediates/library_and_local_jars_jni/release/jni/**/*')
```
chrisfromwork, line 115: @ryantrem let us know if this jni subfolder doesn't get generated on mac. Not sure if its a pc vs mac difference.

Apps/Playground/android/app/build.gradle
```diff
@@ -174,8 +174,7 @@ android {
             def versionCodes = ["armeabi-v7a": 1, "x86": 2, "arm64-v8a": 3, "x86_64": 4]
             def abi = output.getFilter(OutputFile.ABI)
             if (abi != null) {  // null for the universal-debug, universal-release variants
-                output.versionCodeOverride =
-                        versionCodes.get(abi) * 1048576 + defaultConfig.versionCode
+                output.versionCodeOverride = versionCode
```
ryantrem, line 177: This doesn't seem to match upgrade helper.

Modules/@babylonjs/react-native/android/build.gradle
ryantrem: I have 3.5 and this didn't break anything for me, so I think we could just update the comment to say that this is to support the Android gradle plugin 4.0.0+. Please call it "Android gradle plugin" though and not "gradle", because they are versioned differently. We are using gradle version 6.7.

Package/Android/build.gradle
ryantrem: Leave this one as is I think.

Package/gulpfile.js
ryantrem: This worked correctly on Mac.

ryantrem (approved):

Apps/Playground/android/app/build.gradle
chrisfromwork: this should now be fixed

Modules/@babylonjs/react-native/android/build.gradle
chrisfromwork: this should now be fixed

PR #154, opened 2021-02-08
Move windows content into a @babylonjs/react-native-windows package

chrisfromwork:
This pr moves windows content into its own package. This will reduce the @babylonjs/react-native package size for existing customers. It will also allow @babylonjs/react-native-windows to declare its actual peer dependencies (release candidate react-native packages, react-native windows, etc.)


Apps/Playground/package.json
```diff
@@ -15,12 +15,15 @@
     "@babylonjs/core": "^5.0.0-alpha.9",
     "@babylonjs/loaders": "^5.0.0-alpha.9",
     "@babylonjs/react-native": "file:../../Modules/@babylonjs/react-native",
+    "@babylonjs/react-native-windows": "file:../../Modules/@babylonjs/react-native-windows",
     "@react-native-community/slider": "4.0.0-rc.2",
+    "base-64": "^0.1.0",
```
chrisfromwork, line 20: todo remove base-64 and semver

.github/workflows/publish.yml
```diff
@@ -80,6 +71,6 @@ jobs:
         run: |
           npm version --no-git-tag-version ${GITHUB_REF/refs\/tags\//}
```
chrisfromwork, line 72: todo: this version update has not succeeded for me in my fork. something may not be working correctly here

chrisfromwork: this should be fixed by switching to a mac build agent for publishing: https://github.com/chrisfromwork/BabylonReactNative/runs/1858031216?check_suite_focus=true

I was unable to confirm whether multiple packages in the packaging action task broke this version issue, but it seems to work/we'll only be able to test this with the correct permissions provided by the BabylonJS/BabylonReactNative repo.

Modules/@babylonjs/react-native-windows/package.json
```diff
@@ -0,0 +1,52 @@
...
+  "author": {
+    "name": "Your Name",
+    "email": "yourname@email.com"
+  },
+  "homepage": "https://github.com/BabylonJS/BabylonReactNative#readme",
+  "license": "MIT",
+  "licenseFilename": "LICENSE",
+  "readmeFilename": "README.md",
+  "peerDependencies": {
+    "@babylonjs/core": "^5.0.0-alpha.9",
```
ryantrem, line 27: It seems like @babylonjs/react-native should be in this list, right?

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/pch.h
```diff
@@ -30,4 +30,4 @@
 #define NODE_ADDON_API_DISABLE_DEPRECATED
 #endif
 
-#include "../../shared/BabylonNative.h"
\ No newline at end of file
+#include "BabylonNative.h"
```
ryantrem, line 33: Huh, I guess it was just an oversight that I didn't do this for iOS/Android? You didn't change anything to make this possible now where it previously wasn't right?

ryantrem (approved):

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/pch.h
chrisfromwork: the vcxproj did not have the shared folder added as an additional include path. now it does for easier reading

Modules/@babylonjs/react-native-windows/package.json
chrisfromwork: i am not sure how to add @babylonjs/react-native with the correct version given our current tooling. We could add it here as 0.0.1 so that package test evaluates correctly. I don't think this peerDependency version will be automatically changed by our npm packaging setup

ryantrem: I was thinking we would manage it manually. The only part of @babylonjs/react-native that @babylonjs/react-native-windows cares about is the JavaScript right? So technically we'd only have to bump the min version when something in the JS <--> Native communication changes, right?

Does any of the native code (such as headers) get rebundled with @babylonjs/react-native-windows, or is it referenced from @babylonjs/react-native?

ryantrem: This is just a thought though to make it a little more discoverable. We could certainly revisit this in a future PR.

chrisfromwork: @babylonjs/react-native-windows resolves shared headers (BabylonNative.h) from @babylonjs/react-native

I don't think this suggestion of manually managing @babylonjs/react-native version dependencies will scale. It will break local packaging with the 0.0.1 versions. It will also then require prs to update the version prior to every release snap, which is going to be a new pain point for BabylonReactNative contributors

chrisfromwork: i feel like there has to be some other mechanism for declaring a matching version for a dependency

chrisfromwork: oddly enough, if i just add a "@babylonjs/react-native": "version" dependency, i don't see any complaints

PR #155, opened 2021-02-08
Add pointer capture logic to Engine View

chrisfromwork:
We are adding pointer capture logic to the windows engine view to mimic Babylon.js web behavior


Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineView.cpp
```diff
@@ -66,6 +67,14 @@ namespace winrt::BabylonReactNative::implementation {
         const auto position = point.Position();
         const uint32_t x = position.X < 0 ? 0 : static_cast<uint32_t>(position.X);
         const uint32_t y = position.Y < 0 ? 0 : static_cast<uint32_t>(position.Y);
+        const auto pointerId = point.PointerId();
+
+        if (!_inputSource.HasCapture())
+        {
+            _inputSource.SetPointerCapture();
```
ryantrem, line 74: So an input source as a whole has capture, rather than each pointer? For touch input for example, capture applies to all your fingers, not each individual one?

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineView.cpp
```diff
@@ -121,6 +129,14 @@ namespace winrt::BabylonReactNative::implementation {
...
         const uint32_t x = position.X < 0 ? 0 : static_cast<uint32_t>(position.X);
         const uint32_t y = position.Y < 0 ? 0 : static_cast<uint32_t>(position.Y);
+        const auto pointerId = point.PointerId();
+
+        _pressedPointers.erase(pointerId);
+        if (_pressedPointers.size() == 0 &&
+            _inputSource.HasCapture())
+        {
+            _inputSource.ReleasePointerCapture();
+        }
```
ryantrem, line 139: Is it possible to have two different pointer types active at the same time (e.g. mouse and touch)?

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineView.cpp
```diff
@@ -25,16 +25,17 @@ namespace winrt::BabylonReactNative::implementation {
                     static_cast<uint32_t>(Windows::UI::Core::CoreInputDeviceTypes::Mouse) |
                     static_cast<uint32_t>(Windows::UI::Core::CoreInputDeviceTypes::Touch) |
                     static_cast<uint32_t>(Windows::UI::Core::CoreInputDeviceTypes::Pen));
-                auto coreInput = trueThis->CreateCoreIndependentInputSource(deviceTypes);
+                trueThis->_inputSource = trueThis->CreateCoreIndependentInputSource(deviceTypes);
```
ryantrem, line 28: I guess it is not possible to get the `CreateCoreIndependentInputSource` from the event handlers?

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineView.cpp
chrisfromwork: yes. it seems to work differently for CoreIndependentInputSource vs xaml pointers. The pointers obtained here can't be individually captured

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineView.cpp
chrisfromwork: we initially prototyped this to have _inputSource updates run on their own thread outside the threadpool which required access to the _inputSource outside the scope of callbacks. It may be overkill to hold this reference for now, but we probably will move this logic outside the thread pool long term when we can successfully tear down the EngineView/when other graphics life cycle bugs are fixed.

Modules/@babylonjs/react-native/windows/BabylonReactNative/EngineView.cpp
chrisfromwork: I believe so. Input source pointer events seem capable of exposing pen/touch/mouse. I don't personally have a way to test this currently.

ryantrem (approved):

PR #157, opened 2021-02-16
Fix webxrsessionmanager propagation for custom enterXRAsync call

chrisfromwork:
We are currently failing to propagate the WebXRSessionManager for windows enterXRAsync calls. This change should fix things.

### DIFF
--- a/Modules/@babylonjs/react-native/EngineHook.ts
+++ b/Modules/@babylonjs/react-native/EngineHook.ts
@@ -69,10 +69,10 @@ if (Platform.OS == "windows") {
     WebXRExperienceHelper.prototype.enterXRAsync = async function (...args: any[]): Promise<any> {
         // TODO: https://github.com/BabylonJS/BabylonNative/issues/577
         // Windows HMDs require different rendering behaviors than default xr rendering for mobile devices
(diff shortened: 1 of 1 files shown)


Modules/@babylonjs/react-native/EngineHook.ts
```diff
@@ -69,9 +69,12 @@ if (Platform.OS == "windows") {
     WebXRExperienceHelper.prototype.enterXRAsync = async function (...args: any[]): Promise<any> {
         // TODO: https://github.com/BabylonJS/BabylonNative/issues/577
         // Windows HMDs require different rendering behaviors than default xr rendering for mobile devices
-        await originalEnterXRAsync.apply(this, args);
+        const sessionManager = await originalEnterXRAsync.apply(this, args);
         this.scene.clearColor = Color3.Black().toColor4();
         this.scene.autoClear = true;
+        return new Promise<any>((resolve) => {
+            resolve(sessionManager);
```
rgerd, line 76: nit: I would put everything in the promise-- just feels a little strange to have a promise resolve immediately with work that happens outside of its function body thing but inside the function call returning the promise which represents the async work being done to create sessionManager

rgerd (approved):

PR #160, opened 2021-02-21
Direct Assign IDXGISwapChain1 to the ISwapChainPanelNative in Babylon React Native(reworked)

nkari82:
This pull request solves several problems on the Windows version.

1. Problems not being able to redraw when changing JavaScript.
2. Crash occurs when changing the window size in debug mode.
3. By creating a render-only thread, the React native UI is smooth.
4. Destructor not being called in EngineView.

ReactNative has a pre-pull request.
https://github.com/BabylonJS/BabylonNative/pull/606

### DIFF
--- a/Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
+++ b/Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
@@ -53,18 +53,22 @@ namespace winrt::BabylonReactNative::implementation {
 
 	EngineView::~EngineView()
 	{
+		if(_inputLoopWorker)
+			_inputLoopWorker.Cancel();
+
+		if(_renderLoopWorker)
+			_renderLoopWorker.Cancel();
+		
+		if(_resetView.valid())
+			_resetView.wait();	// waiting for bgfx to shutdown..
+
 		_criticalSection.lock();
 		_dxgiOutput = nullptr;
 		_backBufferPtr = nullptr;
 		_swapChain = nullptr;
 		_d3dContext = nullptr;
-		//_d3dDevice = nullptr;		// It will be released on bgfx.
+		_d3dDevice = nullptr;		// It will be released on bgfx.
 		_criticalSection.unlock();
-
-		_inputLoopWorker.Cancel();
-		_renderLoopWorker.Cancel();
-		
-		_resetView.wait();	// waiting for bgfx to shutdown..
 	}
 
 	void EngineView::Reset()
@@ -73,28 +77,25 @@ namespace winrt::BabylonReactNative::implementation {
 
 	void EngineView::RegisterInput()
 	{
-        WorkItemHandler workItemHandler([weakThis{ this->get_weak() }](IAsyncAction const& /* action */)
-        {
-            if (auto trueThis = weakThis.get())
-            {
-                auto deviceTypes = static_cast<CoreInputDeviceTypes>(
-                    static_cast<uint32_t>(Windows::UI::Core::CoreInputDeviceTypes::Mouse) |
-                    static_cast<uint32_t>(Windows::UI::Core::CoreInputDeviceTypes::Touch) |
-                    static_cast<uint32_t>(Windows::UI::Core::CoreInputDeviceTypes::Pen));
-                trueThis->_inputSource = trueThis->CreateCoreIndependentInputSource(deviceTypes);
+		WorkItemHandler workItemHandler([this](IAsyncAction const& /* action */) mutable
+		{
+			auto deviceTypes = static_cast<CoreInputDeviceTypes>(
+				static_cast<uint32_t>(Windows::UI::Core::CoreInputDeviceTypes::Mouse) |
+				static_cast<uint32_t>(Windows::UI::Core::CoreInputDeviceTypes::Touch) |
+				static_cast<uint32_t>(Windows::UI::Core::CoreInputDeviceTypes::Pen));
+			auto coreInput = CreateCoreIndependentInputSource(deviceTypes);
 
-                trueThis->_revokerData.PointerPressedRevoker = trueThis->_inputSource.PointerPressed(winrt::auto_revoke, { trueThis.get(), &EngineView::OnPointerPressed });
-                trueThis->_revokerData.PointerMovedRevoker = trueThis->_inputSource.PointerMoved(winrt::auto_revoke, { trueThis.get(), &EngineView::OnPointerMoved });
-                trueThis->_revokerData.PointerReleasedRevoker = trueThis->_inputSource.PointerReleased(winrt::auto_revoke, { trueThis.get(), &EngineView::OnPointerReleased });
+			auto PointerPressedRevoker = coreInput.PointerPressed(winrt::auto_revoke, { this, &EngineView::OnPointerPressed });
+			auto PointerMovedRevoker = coreInput.PointerMoved(winrt::auto_revoke, { this, &EngineView::OnPointerMoved });
+			auto PointerReleasedRevoker = coreInput.PointerReleased(winrt::auto_revoke, { this, &EngineView::OnPointerReleased });
 
-                trueThis->_inputSource.Dispatcher().ProcessEvents(Windows::UI::Core::CoreProcessEventsOption::ProcessUntilQuit);
-            }
-        });
+			coreInput.Dispatcher().ProcessEvents(Windows::UI::Core::CoreProcessEventsOption::ProcessUntilQuit);
+		});
 
-        // TODO: move to std::thread compared to consuming ThreadPool resources once engine lifecycle bugs are addressed and EngineView's destructor can be successfully invoked.
-        _inputLoopWorker = ThreadPool::RunAsync(workItemHandler, WorkItemPriority::High, WorkItemOptions::TimeSliced);
-    }
-    
+		// TODO: move to std::thread compared to consuming ThreadPool resources once engine lifecycle bugs are addressed and EngineView's destructor can be successfully invoked.
+		_inputLoopWorker = ThreadPool::RunAsync(workItemHandler, WorkItemPriority::High, WorkItemOptions::TimeSliced);
+	}
+
 	void EngineView::RegisterRender()
 	{
 		_criticalSection.lock();
@@ -139,10 +140,6 @@ namespace winrt::BabylonReactNative::implementation {
 		auto swapChainPanel = static_cast<SwapChainPanel>(*this);
 		auto swapChainPanelNative = swapChainPanel.as<ISwapChainPanelNative>().get();;
 		swapChainPanelNative->SetSwapChain(_swapChain.Get());
-
-		// Set render targets to the screen.
-		ID3D11RenderTargetView* const targets[1] = { _backBufferPtr.Get() };
-		_d3dContext->OMSetRenderTargets(1, targets, nullptr);
 	}
 
 	void EngineView::OnPointerPressed(IInspectable const& /*sender*/, PointerEventArgs const& args)
@@ -255,10 +252,6 @@ namespace winrt::BabylonReactNative::implementation {
 			auto swapChainPanel = static_cast<SwapChainPanel>(*this);
 			auto swapChainPanelNative = swapChainPanel.as<ISwapChainPanelNative>().get();;
 			swapChainPanelNative->SetSwapChain(_swapChain.Get());
-
-			// Set render targets to the screen.
-			ID3D11RenderTargetView* const targets[1] = { _backBufferPtr.Get() };
-			_d3dContext->OMSetRenderTargets(1, targets, nullptr);
 		}
 	}
 
@@ -267,7 +260,7 @@ namespace winrt::BabylonReactNative::implementation {
 		critical_section::scoped_lock lock(_criticalSection);
 
 		::Microsoft::WRL::ComPtr<IDXGIDevice3> dxgiDevice;
-		_d3dDevice.As(&dxgiDevice);
+		_d3dDevice->QueryInterface(__uuidof(IDXGIDevice3), &dxgiDevice);
 
 		// Hints to the driver that the app is entering an idle state and that its memory can be used temporarily for other apps.
 		dxgiDevice->Trim();
@@ -275,7 +268,7 @@ namespace winrt::BabylonReactNative::implementation {
 
 	void EngineView::OnResuming()
 	{}
-	
+
 	void EngineView::OnRendering()
 	{
 		Concurrency::critical_section::scoped_lock lock(_criticalSection);
@@ -336,7 +329,7 @@ namespace winrt::BabylonReactNative::implementation {
 			D3D_FEATURE_LEVEL_9_1
 		};
 
-		::Microsoft::WRL::ComPtr<ID3D11Device> device;
+		ID3D11Device* device{ nullptr };
 		::Microsoft::WRL::ComPtr<ID3D11DeviceContext> context;
 		D3D11CreateDevice(
 			nullptr,
@@ -356,7 +349,8 @@ namespace winrt::BabylonReactNative::implementation {
 		);
 
 		// Get D3D11.1 device
-		device.As(&_d3dDevice);
+		if (_d3dDevice) _d3dDevice->Release();
+		device->QueryInterface(__uuidof(ID3D11Device1), (void**)&_d3dDevice);
 
 		// Get D3D11.1 context
 		context.As(&_d3dContext);
@@ -409,8 +403,8 @@ namespace winrt::BabylonReactNative::implementation {
 
 			// Get underlying DXGI Device from D3D Device.
 			::Microsoft::WRL::ComPtr<IDXGIDevice1> dxgiDevice;
-			_d3dDevice.As(&dxgiDevice);
-
+			_d3dDevice->QueryInterface(__uuidof(IDXGIDevice1), &dxgiDevice);
+			
 			// Get adapter.
 			::Microsoft::WRL::ComPtr<IDXGIAdapter> dxgiAdapter;
 			dxgiDevice->GetAdapter(&dxgiAdapter);
@@ -424,7 +418,7 @@ namespace winrt::BabylonReactNative::implementation {
 			// Create swap chain.
 			::Microsoft::WRL::ComPtr<IDXGISwapChain1> swapChain;
 			dxgiFactory->CreateSwapChainForComposition(
-				_d3dDevice.Get(),
+				_d3dDevice,
 				&scd,
 				nullptr,
 				&swapChain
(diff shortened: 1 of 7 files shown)


Modules/@babylonjs/react-native/shared/BabylonNative.cpp
```diff
@@ -123,7 +124,8 @@ namespace Babylon
                 throw std::runtime_error{ "RenderView can only be called when automatic rendering is disabled." };
             }
 
-            m_graphics->RenderCurrentFrame();
+            if(m_graphics)
+				m_graphics->RenderCurrentFrame();
```
chrisfromwork, line 128: >				m_g [](start = 0, length = 7)

nit: we should probably do a pass and make sure all tabs are replaced with spaces

chrisfromwork: the selected text for this comment looks a bit strange. The white space leading this m_graphics->RenderCurrentFrame() uses tabs compared to spaces, which is the actual thing we will want to update

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
```diff
@@ -2,173 +2,452 @@
...
+	void EngineView::OnRendering()
+	{
+		Concurrency::critical_section::scoped_lock lock(s_criticalSection);
+
+		if (!_backBufferPtr)
+			return;
+
+		// Clear the back buffer and depth stencil view.
+		float clearColor[4]{ 0.392156899f, 0.584313750f, 0.929411829f, 1.000000000f };
+		s_d3dContext->ClearRenderTargetView(_backBufferPtr.Get(), clearColor);
```
chrisfromwork, line 279: is this the blue background that can be seen when initializing? Would we prefer to have this as black?

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
```diff
@@ -2,173 +2,452 @@
...
+
+	void EngineView::OnCompositionScaleChanged(SwapChainPanel const& sender, IInspectable const& /*object*/)
+	{
+		if (_compositionScaleX != sender.CompositionScaleX() || _compositionScaleY != sender.CompositionScaleY())
+		{
+			critical_section::scoped_lock lock(s_criticalSection);
+
+			// Store values so they can be accessed from a background thread.
+			_compositionScaleX = sender.CompositionScaleX();
+			_compositionScaleY = sender.CompositionScaleY();
```
chrisfromwork, line 245: When testing this locally, have you ever seen a composition scale != 1? When testing this out on my end I see the correct visual scaling when this composition scale has a value of 1. After initialization this value gets set to 2, which then causes the content to display incorrectly.

chrisfromwork:
Screenshot (76)
I am seeing this sort of rendering behavior when composition scale == 2. I am not yet sure if we need to do something differently to recreate the device resources or if we need to avoid applying the composition scale all together.

chrisfromwork:
regardless, this is an awesome change. It seems like now toggling the engine view does not result in any crashes. :)

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
```diff
@@ -2,173 +2,452 @@
...
+		s_criticalSection.unlock();
+		
+		// Calculate the updated frame and render once per vertical blanking interval
+		WorkItemHandler workItemHandler([this](IAsyncAction const& action) mutable
+		{
+			//{
+			//	critical_section::scoped_lock lock(s_criticalSection);
+			//	Babylon::EnableView();
+			//}
+
```
chrisfromwork, line 113: should we file a follow up issue to look into reenabling this commented out code?

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
```diff
@@ -2,173 +2,452 @@
...
+		critical_section::scoped_lock lock(s_criticalSection);
+
+		const auto size = args.NewSize();
+		_width = std::max(size.Width, 1.0f);
+		_height = std::max(size.Height, 1.0f);
+
+		CreateSizeDependentResources();
+
+		auto swapChainPanel = static_cast<SwapChainPanel>(*this);
+		auto swapChainPanelNative = swapChainPanel.as<ISwapChainPanelNative>().get();;
```
chrisfromwork, line 139: >; [](start = 79, length = 1)

nit: remove extra ';'

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
chrisfromwork: For what its worth, the swap chain panel size changed event already reports width/height in pixels.

chrisfromwork: it seems like there may be issues here related to: https://github.com/BabylonJS/BabylonNative/pull/582

When i compare this to existing behaviors, this composition scaling change now occurs after babylon.js is initialized. Previously, the composition scale was non one before initialization, which seemed to successfully get relayed to the engine. I'll continue to see whether we can figure out how to fix this.

chrisfromwork: it seems like we may need to call Graphics::SetHardwareScalingLevel to submit this value

chrisfromwork:
@rgerd @tymlipari @ryantrem @syntheticmagus @bghgary @Drigax can you take a look at this review?

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
```diff
@@ -142,7 +144,66 @@ namespace Babylon
...
+                return;
+            
+            if (m_autoRender)
+            {
+                m_graphics->EnableRendering();
+            }
+            else
+            {
+                JSGuard guard(m_jsDispatcher);
+                m_graphics->EnableRendering();
```
ryantrem, line 187: So basically this is going to prevent the JS thread from doing anything during a call to `EnableRendering`? I think this won't be necessary after @bghgary's in flight changes that remove auto rendering and makes graphics threading much more robust. We'll probably need to wait for that change to wrap up and then see how this part of the change would apply (if at all).

Modules/@babylonjs/react-native/shared/BabylonNative.h
```diff
@@ -8,8 +8,10 @@ namespace Babylon
 
     void Initialize(facebook::jsi::Runtime& jsiRuntime, Dispatcher jsDispatcher, bool autoRender = true);
     void Deinitialize();
-    void UpdateView(void* windowPtr, size_t width, size_t height, void* windowTypePtr = nullptr);
+    void UpdateView(void* windowPtr, size_t width, size_t height, void* windowTypePtr = nullptr, void* contextPtr = nullptr, void* backBufferPtr = nullptr);
     void RenderView();
+    void EnableView();
+    void DisableView();
```
ryantrem, line 14: Are these actually used? Maybe I just missed it, not sure!

nkari82: This code is not used.

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
nkari82: Blocks JavaScript threads of Babylon js mapped to bgfx.

I don't actually use it. I wasted time safely shutting down and recreating bgfx. Even if I shut it down safely, it crashes by bgfx related call mapped with babylonjs before bgfx is created again (so it doesn't shut down bgfx).

chrisfromwork: thanks for the context @nkari82!

chrisfromwork:
I've been spending some more time looking at this review. If we don't end up listening/acting on composition scale changes for the swap chain panel well consistently draw content with the correct size. However, it seems like something about how this is all set up prevents anti aliasing from working correctly:

Screenshot (86)
Screenshot (87)

TBD why anti aliasing doesn't seem to be working. I'll provide more information here as we determine whats going on. @nkari82  if you have any initial thoughts on why this may be happening definitely let us know. :)

rgerd:
If you want to have antialiasing, you could do it similarly to how bgfx (BabylonNative's rendering backend) does it:

[They create a texture outside of the swapchain that supports MSAA](https://github.com/bkaradzic/bgfx/blob/master/src/renderer_d3d11.cpp#L1083-L1101)

[They use that MSAA texture in their render target view](https://github.com/bkaradzic/bgfx/blob/master/src/renderer_d3d11.cpp#L2221)

([Which would be analogous to this line in your PR](https://github.com/BabylonJS/BabylonReactNative/pull/160/files#diff-59eecabc40d052e377bb76ad16e57d580211d2c27a11b60f471c7687cb10ea29R431))

[You'll want to be sure you create a multisampled depth/stencil buffer as well, because I'm pretty sure this code won't run with multisampling in the sample description when creating the depth/stencil buffer for you](https://github.com/bkaradzic/bgfx/blob/master/src/renderer_d3d11.cpp#L2231-L2249)

[And then when they're finished they resolve the MSAA render target into their swapchain color buffer](https://github.com/bkaradzic/bgfx/blob/master/src/renderer_d3d11.cpp#L6578-L6579), which I assume you would do after the call to `RenderView()`.

--

[Here's a description of MSAA for those who haven't been exposed to it yet](https://learnopengl.com/Advanced-OpenGL/Anti-Aliasing)

[And a parallel description on how to do multisampling in a generic UWP app](https://docs.microsoft.com/en-us/windows/uwp/gaming/multisampling--multi-sample-anti-aliasing--in-windows-store-apps)

--

None of this is tested, just an educated guess. I could spend some time proofing it out since it should be pretty straightforward to attempt.

nkari82:
here! msaa
msaa

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
```diff
@@ -163,12 +237,279 @@ namespace winrt::BabylonReactNative::implementation {
...
+        // present
+        DXGI_PRESENT_PARAMETERS parameters = { 0 };
+        parameters.DirtyRectsCount = 0;
+        parameters.pDirtyRects = nullptr;
+        parameters.pScrollRect = nullptr;
+        parameters.pScrollOffset = nullptr;
+
+        if (_sampleDesc.Count > 1)
+        {
+            ::Microsoft::WRL::ComPtr<ID3D11Texture2D> originBackBuffer, msaaTexutePtr;
```
rgerd, line 300: `msaaTexturePtr`

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.h
```diff
@@ -1,37 +1,59 @@
...
-            winrt::Windows::UI::Core::CoreIndependentInputSource::PointerReleased_revoker PointerReleasedRevoker{};
-            winrt::Windows::UI::Xaml::Media::CompositionTarget::Rendering_revoker RenderingRevoker{};
+            winrt::Windows::UI::Xaml::IApplication::Suspending_revoker SuspendingRevoker{};
+            winrt::Windows::UI::Xaml::IApplication::Resuming_revoker ResumingRevoker{};
         };
         RevokerData _revokerData{};
+
+        DXGI_SAMPLE_DESC _sampleDesc{ 1, 0 };
+        ::Microsoft::WRL::ComPtr<IDXGISwapChain2>_swapChain;
+        ::Microsoft::WRL::ComPtr <ID3D11RenderTargetView> _backBufferPtr;
```
rgerd, line 53:
```suggestion
        ::Microsoft::WRL::ComPtr<ID3D11RenderTargetView> _backBufferPtr;
```

(Same on below line)

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
```diff
@@ -142,7 +144,7 @@ namespace Babylon
                 }
             });
         }
-
+
```
rgerd, line 147: (nit, remove whitespace to avoid use of tabs as well as file diff churn)
```suggestion

```

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
```diff
@@ -278,7 +280,7 @@ namespace Babylon
             throw std::runtime_error{ "RenderView must not be called before Initialize." };
         }
     }
-
+
```
rgerd, line 283: (nit, remove whitespace to avoid use of tabs as well as file diff churn)
```suggestion

```

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
```diff
@@ -163,12 +237,279 @@ namespace winrt::BabylonReactNative::implementation {
...
+            {
+                // If the device was removed for any reason, a new device and swap chain will need to be created.
+                OnDeviceLost();
+                return;
+
+            }
+        }
+        else // Otherwise, create a new one
+        {
+            _sampleDesc = s_msaa[static_cast<uint32_t>(MSAA::X4)];  // MSAA::X4 Default Setting in BabylonNative
```
rgerd, line 436: Seems like this is fine for now, but it might be worth filing an issue to eventually make this externally configurable, since I could see BNRN developers potentially wanting control over the MSAA level to balance performance / video memory / visual quality.

chrisfromwork:
@rgerd and @nkari82 although this logic works, we may want to assess whether we can introduce the fixes into bgfx. This change in its current form seems like it is now reimplementing a good amount of logic already in bgfx to support MSAA.

@nkari82 it seems like a few of the underlying crashes are related to the SwapChainPanel still holding a reference to the underlying swap chain when its released by bgfx (this happens during both resize operations and engine tear down). Are there any other clear issues with bgfx in regards to hot reload that this review is addressing?

nkari82:
@chrisfromwork
There are two problems.
1. Crash in bgfx in swapchain when resizing
2. 3d object is not visible during hot reloading. (I modified and tested bgfx regarding the swap chain.)

I don't want to fix bgfx if there is a way to get around the problem.

The amount of code is still simple, so it's okay, but if the code grows more, I think a way to fix bgfx would be okay.

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
nkari82: that's right

chrisfromwork:
FWIW, @nkari82 I am going to work to get this change checked in as well: https://github.com/BabylonJS/bgfx/pull/15

nkari82:
@chrisfromwork

I looked at the changes in bgfx.

https://github.com/bkaradzic/bgfx/blob/master/src/dxgi.cpp#L592

What do you think about "swapChainPanelNative->SetSwapChain(NULL)" behavior at this location when a destructor is called from EngineView? Then I don't think I need a "removeSwapChain" :)

chrisfromwork:
> @chrisfromwork
> 
> I looked at the changes in bgfx.
> 
> https://github.com/bkaradzic/bgfx/blob/master/src/dxgi.cpp#L592
> 
> What do you think about "swapChainPanelNative->SetSwapChain(NULL)" behavior at this location when a destructor and OnSizeChanged is called from EngineView? Then I don't think I need a "removeSwapChain" :)

we can definitely do this

> @chrisfromwork
> 
> I looked at the changes in bgfx.
> 
> https://github.com/bkaradzic/bgfx/blob/master/src/dxgi.cpp#L592
> 
> What do you think about "swapChainPanelNative->SetSwapChain(NULL)" behavior at this location when a destructor and OnSizeChanged is called from EngineView? Then I don't think I need a "removeSwapChain" :)

After looking at this on my end there were a few bugs I saw:
* bgfx did not remove the swap chain from the swap chain panel when destroying it (this is now fixed)
* bgfx did not update the SwapChainPanel reference in its renderer after initialization. This made it so destroying/creating new EngineViews did not work correctly (this is now fixed)
* the react-native-windows Abi safe jsi runtime is currently destroyed prior to cleaning up the underlying jsi objects/jsi runtime. This results in javascript logic continuing to execute when things are in a bad state, which leads to crashes. This will be fixed with this pr: https://github.com/microsoft/react-native-windows/pull/7356

We could do this SetSwapChain call on our end, but it probably isn't necessary now that bgfx has fixes. That all going said, these different fixes I have introduced avoid crashes but have resulted in white panel flashing when the swap chain panel is changed. There is still room for improvement here specifically related to continuing to show babylon.js content during resize/fast refresh scenarios

nkari82:
> > @chrisfromwork
> > I looked at the changes in bgfx.
> > https://github.com/bkaradzic/bgfx/blob/master/src/dxgi.cpp#L592
> > What do you think about "swapChainPanelNative->SetSwapChain(NULL)" behavior at this location when a destructor and OnSizeChanged is called from EngineView? Then I don't think I need a "removeSwapChain" :)
> 
> we can definitely do this
> 
> > @chrisfromwork
> > I looked at the changes in bgfx.
> > https://github.com/bkaradzic/bgfx/blob/master/src/dxgi.cpp#L592
> > What do you think about "swapChainPanelNative->SetSwapChain(NULL)" behavior at this location when a destructor and OnSizeChanged is called from EngineView? Then I don't think I need a "removeSwapChain" :)
> 
> After looking at this on my end there were a few bugs I saw:
> 
> * bgfx did not remove the swap chain from the swap chain panel when destroying it (this is now fixed)
> * bgfx did not update the SwapChainPanel reference in its renderer after initialization. This made it so destroying/creating new EngineViews did not work correctly (this is now fixed)
> * the react-native-windows Abi safe jsi runtime is currently destroyed prior to cleaning up the underlying jsi objects/jsi runtime. This results in javascript logic continuing to execute when things are in a bad state, which leads to crashes. This will be fixed with this pr: [microsoft/react-native-windows#7356](https://github.com/microsoft/react-native-windows/pull/7356)
> 
> We could do this SetSwapChain call on our end, but it probably isn't necessary now that bgfx has fixes. That all going said, these different fixes I have introduced avoid crashes but have resulted in white panel flashing when the swap chain panel is changed. There is still room for improvement here specifically related to continuing to show babylon.js content during resize/fast refresh scenarios

thanks for sharering~

chrisfromwork:
@nkari82 at this time we think we have a working fast refresh experience for react-native-windows. There are still some race conditions around loading textures in BabylonNative when tearing down the jsi runtime, but we will be working through those as well. Are there other items that you feel are still needed for react-native-windows fast refresh?

nkari82:
> @nkari82 at this time we think we have a working fast refresh experience for react-native-windows. There are still some race conditions around loading textures in BabylonNative when tearing down the jsi runtime, but we will be working through those as well. Are there other items that you feel are still needed for react-native-windows fast refresh?

chrisfromwork:
@nkari82 that's super cool that you have a typescript bgfx wrapper working! 

I haven't spent a ton of time looking into custom asset loaders/shader compilers/texture converters. But we definitely could add these to the issues list if you want to file feature requests. :)

PR #164, opened 2021-02-25
Fix release builds and update react-native-windows dependencies

chrisfromwork:
This review fixes the BabylonReactNative.vcxproj to reference lib directories over project vcxproj files. This seems to fix issues with linking where debug libs can be unintentionally referenced.

This review also moves @babylonjs/react-native-windows to use more up to date react-native-windows dependencies


Apps/Playground/package.json
```diff
@@ -19,9 +19,9 @@
     "@react-native-community/slider": "4.0.0-rc.2",
     "logkitty": "^0.7.1",
     "react": "^17.0.1",
-    "react-native": "0.64.0-rc.2",
+    "react-native": "^0.64.0-rc.0",
```
ryantrem, line 22: This older version is new enough?

ryantrem (commented):
With the way it was (project references), when you build the top level `BabylonReactNative.vcxproj` that would also build all the referenced projects, but that's not longer the case. What is the process to build all those dependencies? Do you just build a cmake generated sln or something?

chrisfromwork:
> With the way it was (project references), when you build the top level `BabylonReactNative.vcxproj` that would also build all the referenced projects, but that's not longer the case. What is the process to build all those dependencies? Do you just build a cmake generated sln or something?

This was the original expectation when referencing the vcxproj files, but it did not always seem to be the case. File updates/rebuilds were not getting triggered for me but dependency references were working with the current project setup. Right now we rebuild BabylonNative libs by running the gulp command or by building the generated ReactNativeBabylon.sln from cmake

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/BabylonReactNative.vcxproj
```diff
@@ -125,21 +159,8 @@
...
+      <AdditionalDependencies>arcana.lib;astc.lib;astc-codec.lib;BabylonNative.lib;bgfx.lib;bimg.lib;bx.lib;edtaa3.lib;etc1.lib;etc2.lib;GenericCodeGend.lib;glslangd.lib;Graphics.lib;iqa.lib;jsi.lib;JsRuntime.lib;MachineIndependentd.lib;napi.lib;NativeCapture.lib;NativeEngine.lib;NativeInput.lib;NativeXr.lib;nvtt.lib;OGLCompilerd.lib;openxr_loader.lib;OSDependentd.lib;pvrtc.lib;spirv-cross-core.lib;spirv-cross-glsl.lib;spirv-cross-hlsl.lib;SPIRVd.lib;squish.lib;UrlLib.lib;Window.lib;XMLHttpRequest.lib;xr.lib;WindowsApp.lib;%(AdditionalDependencies);</AdditionalDependencies>
```
ryantrem, line 162: Should we put this list of libs in a property so we don’t have to define it multiple times?

chrisfromwork: right now its defined twice because some libs have a 'd' suffix for debug names. we could do this if you still think it will be helpful, but it would require multiple variables

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/BabylonReactNative.vcxproj
```diff
@@ -115,6 +124,31 @@
       <SubSystem>Console</SubSystem>
       <GenerateWindowsMetadata>true</GenerateWindowsMetadata>
       <ModuleDefinitionFile>BabylonReactNative.def</ModuleDefinitionFile>
+      <!-- Library directories when building in repo -->
+      <!-- Note: you can add BabylonNative component build folders here as needed -->
+      <AdditionalLibraryDirectories Condition="'$(BabylonNativeBuildDir)' != ''">$(BabylonNativeBuildDir)\Dependencies\arcana.cpp\$(Configuration);%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
```
tymlipari, line 129: >AdditionalLibraryDirectories [](start = 7, length = 28)

nit: MSBuild supports multi-line properties, you don't need to duplicate this over and over. Just do something like:

```
<AdditionalLibraryDirectories Condition="'$(BabylonNativeBuildDir)' != ''">
    $(BabylonNativeBuildDir)\Dependencies\bgfx.cmake\$(Configuration);
    $(BabylonNativeBuildDir)\Dependencies\glslang\glslang\$(Configuration);
    ...
    %(AdditionalLibraryDirectories)
</AdditionalLibraryDirectories>
```

chrisfromwork: i feel like i had originally tried this but hit issues. I'll look into this again

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/BabylonReactNative.vcxproj
chrisfromwork: i've updated this to improve legibility, but there are two locations to declare libs still (one for debug, one for release)

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/BabylonReactNative.vcxproj
chrisfromwork: this should now be updated

Apps/Playground/package.json
ryantrem: If your app has already installed react-native@0.64.0-rc.0 and that's what is in the package.json.lock file, then it won't automatically update to something newer, so does @babylonjs/react-native-windows actually work with 0.64.0-rc.0?

ryantrem (approved):

Apps/Playground/package.json
chrisfromwork: I don't think it does. Would peer dependency resolution be able to pick up these items? Seems like maybe I should specify a greater than clause as well

PR #167, opened 2021-03-01
Enable all build flavors for UWP github PR actions

chrisfromwork:
We noticed recently that the release flavor was not building correctly. To try and avoid this in the future, this pr adds tasks to build all UWP flavors for prs.

Note: UWP release flavors may fail until the following pr is checked in: https://github.com/BabylonJS/BabylonReactNative/pull/164


.github/workflows/pr.yml
```diff
@@ -63,6 +63,260 @@ jobs:
...
+        run: npx gulp buildUWPx86Debug
+        working-directory: ./Package
+      - name: Gulp NuGet Restore Playground
+        run: npx gulp nugetRestoreUWPPlayground
+        working-directory: ./Package
+      - name: Gulp Build x86 Debug Playground (Windows)
+        run: npx gulp buildUWPPlaygroundx86Debug
+        working-directory: ./Package
+
+  build-windows-x86-release:
```
ryantrem, line 79: These all look like they are basically clones with configuration variations only, is that right? If so, can we leverage the GitHub workflow "matrix strategy" stuff? https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

chrisfromwork: yes. i will look into this

chrisfromwork: we now use a matrix for setup. Please let me know whether you are also wanting to reduce the number of gulp tasks/if you know of examples for providing arguments to gulp tasks

ryantrem (approved):

chrisfromwork:
resetting android build

PR #173, opened 2021-03-12
Create dedicated XR view

ryantrem:
This change updates Babylon Native to the latest and provides a dedicated XR view (corresponds with https://github.com/BabylonJS/BabylonNative/pull/631).

The Babylon Native submodule will need to be updated to master after the Babylon Native pull request mentioned above is completed.


Modules/@babylonjs/react-native/android/src/main/cpp/BabylonNativeInterop.cpp
```diff
@@ -82,6 +82,21 @@ extern "C" JNIEXPORT void JNICALL Java_com_babylonreactnative_BabylonNativeInter
     Babylon::ResetView();
 }
 
+extern "C" JNIEXPORT void JNICALL Java_com_babylonreactnative_BabylonNativeInterop_00024BabylonNative_updateXRView(JNIEnv* env, jclass obj, jobject surface)
+{
+    ANativeWindow *windowPtr{};
```
bghgary, line 87: nit: 
```suggestion
    ANativeWindow* windowPtr{};
```

bghgary (approved):

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
```diff
@@ -270,4 +285,22 @@ namespace Babylon
...
+        }
+
+        return false;
+    }
+
+    void UpdateXRView(void* windowPtr)
+    {
+        if (auto nativeModule{ g_nativeModule.lock() })
+        {
+            nativeModule->UpdateXRView(windowPtr);
```
chrisfromwork, line 303: Is this a no-op on windows? Do we think that any of these changes will make oculus quest (openxr + android) scenarios more difficult to implement?

ryantrem: It's basically an "optional" api, in the same way that we used to pass the main windowPtr to the xr.cpp layer, but it was ignored in the OpenXR implementation. So on Windows for now we can just not use this API to pass along a window pointer. Conceptually the difference is between when XR needs to render to some render surface that is managed by the caller vs. a render surface that is managed by xr.cpp itself (such as on an HMD where we are fully immersive and render "full screen"). We probably can do more to rationalize these different scenarios in the NativeXr API, but I think we can do that in a future change.

Alex-MSFT (approved):

PR #174, opened 2021-03-12
Enable fast refresh for UWP scenarios

chrisfromwork:
react-native-windows 0.64.0-preview.15 contains jsi runtime fixes that allow fast refresh to work for UWP scenarios. This change updates package dependencies to 0.64.0-preview.15. UWP app instance settings have also been updated to use fast refresh for debug builds.

note: there are still issues around fast refresh when in an immersive xr session


Apps/Playground/windows/Playground/App.cpp
```diff
@@ -25,14 +25,14 @@ using namespace Windows::ApplicationModel;
...
 {
 #if BUNDLE
-    // Note: add <UseBundle>true</UseBundle> property to application's ReactNativeWindowsProps PropertyGroup to force BUNDLE to true
     JavaScriptBundleFile(L"index.windows");
     InstanceSettings().UseWebDebugger(false);
     InstanceSettings().UseFastRefresh(false);
 #else
     JavaScriptMainModuleName(L"index");
-    InstanceSettings().UseWebDebugger(true);
+    InstanceSettings().UseWebDebugger(false); // BabylonReactNative accesses the jsi runtime, which isn't possible with the web debugger
```
ryantrem, line 33: On other platforms, web debugger (remote debugging) is enabled/disabled dynamically from the app's debug/developer menu. Is this different on Windows? It would be annoying if all UWP RN apps have to change this configuration in order to use Babylon React Native.

ryantrem (approved):

Apps/Playground/windows/Playground/App.cpp
chrisfromwork: it sounds like it may be possible to set this dynamically, but i'm still trying to track down documentation on how. For the time being, this hardcoding of the ip address is a valid workaround/has been allowing me to test things on hololens 2

PR #196, opened 2021-04-09
Add a helper header and update gulp to package needed headers for using an ARCore xr context

chrisfromwork:
This review introduces a new XrContextHelperARCore.h header that helps developers access the arcore xr context through the global xr component.

This review also updates gulp scripts to include IXrContextARCore.h and XrContextHelperARCore.h in the distributed @babylonjs/react-native npm package.

Note: we don't currently have a great way to test this logic change. We will validate and convert this pr to non-draft while working on Azure Spatial Anchors android support


rgerd (approved):

Modules/@babylonjs/react-native/android/include/XrContextHelperARCore.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#if __has_include("jsi/jsi.h")
+#include "jsi/jsi.h"
+#include "IXrContextARCore.h"
+
+namespace BabylonReactNative
```
ryantrem, line 7: It might make sense to follow the namespace naming scheme of Babylon Native here. So something like Babylon::Plugins::NativeXr as a clear indication that this is tied to that plugin.

Modules/@babylonjs/react-native/android/include/XrContextHelperARCore.h
```diff
@@ -0,0 +1,36 @@
...
+            return false;
+        }
+
+        auto nativeXr{ navigator.getProperty(jsiRuntime, "xr").asObject(jsiRuntime) };
+        if (!nativeXr.hasProperty(jsiRuntime, "nativeXrContext") ||
+            !nativeXr.hasProperty(jsiRuntime, "nativeXrContextType") ||
+            nativeXr.getProperty(jsiRuntime, "nativeXrContextType").asString(jsiRuntime).utf8(jsiRuntime) != "ARCore")
+        {
+            return false;
+        }
```
ryantrem, line 29: Should we move this to a shared header? Is it worth it? Seems like we're starting to get some duplication across platforms and features.

Modules/@babylonjs/react-native/android/include/XrContextHelperARCore.h
chrisfromwork: this should now be updated

PR #200, opened 2021-04-15
Move BabylonNative UWP build out of submodule

rgerd:
I was running into possible path length issues when making the UWP projects, this shortens the paths and moves the builds into a Build/ folder, which gets ignored by the .gitignore


chrisfromwork:
@rgerd I'm going to sign off, but please confirm that the package test 0.64 app still works with this change (this will help ensure publishing @babylonjs/react-native-windows isn't regressed and will run the gulp package tasks)

chrisfromwork (approved):

PR #204, opened 2021-04-20
Add check to make sure folks are using npm 6.13 with playground

chrisfromwork:

1. The readme now contains a known issues section
2. running npm install in Apps\Playground now assesses the npm version folks are using to make sure they aren't trying to build with npm 7+

### DIFF
--- a/Apps/Playground/scripts/postinstall.js
+++ b/Apps/Playground/scripts/postinstall.js
@@ -15,7 +15,7 @@ else {
     const version = parts[1].trim();
 
     // Check if using non 6.13.* npm version
-    if (!version.match(/6\.13\..*/g)) {
+    if (!version.match(/6\.\d+\.\d+/g)) {
         throw `Error: BabylonReactNative Playground development requires npm version 6.13.*, Your current npm version is ${version}. Run npm install -g npm@6.13 to update your npm version.`;
     }
 }
--- a/Apps/Playground/scripts/postinstall.js
+++ b/Apps/Playground/scripts/postinstall.js
@@ -8,7 +8,6 @@ if (!output.includes("npm@")) {
 
 if (!output.includes("npm@")) {
     shelljs.echo("local npm version not found");
-
 }
 else {
     const parts = output.split("npm@");
--- a/Apps/Playground/scripts/postinstall.js
+++ b/Apps/Playground/scripts/postinstall.js
@@ -14,7 +14,7 @@ else {
     const parts = output.split("npm@");
     const version = parts[1].trim();
 
-    // Check if using non 6.13.* npm version
+    // Check if using non 6.*.* npm version
     if (!version.match(/6\.\d+\.\d+/g)) {
         throw `Error: BabylonReactNative Playground development requires npm version 6.13.*, Your current npm version is ${version}. Run npm install -g npm@6.13 to update your npm version.`;
     }
--- a/Apps/Playground/scripts/postinstall.js
+++ b/Apps/Playground/scripts/postinstall.js
@@ -1,10 +1,21 @@
 const shelljs = require('shelljs');
-const output = shelljs.exec('npm list -g npm');
-const parts = output.split("npm@");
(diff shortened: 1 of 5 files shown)


Package/gulpfile.js
```diff
@@ -439,6 +439,18 @@ const createPackageUWP = async () => {
   exec('npm pack', 'Assembled-Windows');
 }
 
+const checkNPMVersion = async () => {
+  const output = shelljs.exec('npm list -g npm');
+  const parts = output.split("npm@");
+  const version = parts[1].trim();
+  
+  // Check if using npm version 7+
+  if (version.charAt(0) == '7')
```
chrisfromwork, line 448: not sure if folks would prefer something more intelligent here.

Apps/Playground/scripts/postinstall.js
```diff
@@ -0,0 +1,10 @@
+const shelljs = require('shelljs');
+const output = shelljs.exec('npm list -g npm');
+const parts = output.split("npm@");
+const version = parts[1].trim();
+
+// Check if using npm version 7+
+if (version.charAt(0) == '7')
```
rgerd, line 7: You could also do something like `!!version.match(/6\.13\..*/g)`, or import semver

rgerd (approved):

Apps/Playground/scripts/postinstall.js
chrisfromwork: this should now be updated

Apps/Playground/scripts/postinstall.js
```diff
@@ -0,0 +1,21 @@
...
+let output = shelljs.exec('npm list -g npm');
+
+if (!output.includes("npm@")) {
+    shelljs.echo("global npm install not found");
+    output = shelljs.exec('npm list npm');
+}
+
+if (!output.includes("npm@")) {
+    shelljs.echo("local npm version not found");
+
```
rgerd, line 11: nit: line

rgerd:
You could also in theory do something like:
```javascript
"preinstall": "npm --version | scripts/preinstall.js"
```

```javascript
function getStandardInput() {
    return new Promise((resolve, reject) => {
        const stdin = process.stdin;
        let data = '';
        stdin.setEncoding('utf8');
        stdin.on('data', (chunk) => data += chunk);
        stdin.on('end', () => resolve(data));
        stdin.on('error', reject);
    });
}

function checkNPMVersion(npmOutput) {
    const version = npmOutput.trim();

    // Check if using non 6.13.* npm version
    if (!npmOutput.match(/6\.13\..*/g)) {
        throw `
Error: BabylonReactNative Playground development requires npm version 6.13.*.
Your current npm version is ${version}. Run npm install -g npm@6.13 to update your npm version.
            `;
    }
}

getStandardInput().then(checkNPMVersion).catch(console.error);
```
(Source for piping: https://wellingguzman.com/notes/node-pipe-input)

Apps/Playground/scripts/postinstall.js
```diff
@@ -0,0 +1,6 @@
+const shelljs = require('shelljs');
+const version = shelljs.exec('npm --version');
+
+if (!version.trim().match(/6\.\d+\.\d+/g)) {
```
ryantrem, line 4: Sorry I know there has already been some discussion on this and I meant to comment on it earlier but forgot... you could also consider using the `semver` package, which would make this code a bit more readable. Something like:
```js
const semver = require('semver');
const shelljs = require('shelljs');
const version = semver.clean(shelljs.exec('npm --version'));

if (!semver.lt(version, '7')) {
    throw ...
}
```
Or you might need to do `semver.lt(version, semver.coerce('7'))`, not sure. Anyway up to you, but the `semver` package is all about dealing with versions and version comparisons.

chrisfromwork: sounds good. going to go with the regex for now, we can update this in the future as needed!

PR #207, opened 2021-04-26
Add github action to run when BabylonNative is updated

chrisfromwork:
This github action will check out the most recent commit in babylonnative and attempt a build. It should allow us to monitor incoming changes.

once this change has been proven out, we can have additional ci to look into merging the babylonnative master branch on successful build.

Note: we will need the following change for these jobs to start running: https://github.com/BabylonJS/BabylonNative/pull/714

### DIFF
--- a/.github/workflows/bn_master_commit.yml
+++ b/.github/workflows/bn_master_commit.yml
@@ -75,7 +75,7 @@ jobs:
         run: npm install
         working-directory: ./Package
       - name: Git (Update to BabylonNative master)
-        run: npx gulp initializeSubmodulesMostRecentBabylonNative
+        run: npx gulp initializeSubmodulesMostRecentBabylonNativeWindowsAgent
         working-directory: ./Package
       - name: Gulp Setup Project ${{ matrix.platform }} (Windows)
         run: npx gulp makeUWPProject${{ matrix.platform }}
--- a/.github/workflows/bn_master_commit.yml
+++ b/.github/workflows/bn_master_commit.yml
@@ -1,0 +1,99 @@
+name: BabylonNative master branch update
+
+on:
+  repository_dispatch:
+    types: [babylonnative-master-update]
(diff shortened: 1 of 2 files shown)


.github/workflows/bn_master_commit.yml
```diff
@@ -0,0 +1,91 @@
+name: BabylonNative master branch update
+
+on:
+  repository_dispatch:
+    types: [babylonnative-master-update]
+
+jobs:
```
rgerd, line 7: nit: Feels like you could use a template for this instead of needing to copy job definitions: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops

chrisfromwork: im not sure if azure dev ops templates can be used for github actions. We also have different submodule initialization steps in this definitions

rgerd: It seems like pr.yml and bn_master_commit.yml are similar enough to combine-- you could avoid the `initializeSubmodulesMostRecentBabylonNative` with an if on the step using an input to the composite action...

I'm fine with punting the cleanup, just wanted to point it out

ryantrem: If it is possible to not duplicate all this that would be great. I hadn't heard of a way to do that with GitHub actions, but maybe the composite thing you mentioned is a solution?

chrisfromwork: it seems like this composite solution is good for defining actions in their own repos. The description doesn't show how to do this for repo specific action reuse. I don't think these actions warrant their own repo

chrisfromwork: definitely agree with the sentiment of not wanting to rewrite the same action content

rgerd (approved):

PR #227, opened 2021-05-13
Decreasing screenshot jpeg quality to 80%

rarattay:
Decreasing the screenshot jpeg quality for both iOS and Android to 80% to optimize the returned image file size.  

- 80% jpeg quality is considered to be the recommended quality and the lowest quality setting that has minimal to no visible difference to 100%. 
- Tested on both Android and iOS 
- Attached two screenshots taken on Android at 80% and 100%. 100% = 105kb, 80% = 40kb

100% Quality
100OnDevice

80% Quality
80OnDevice

### DIFF
--- a/Modules/@babylonjs/react-native/android/src/main/java/com/babylonreactnative/EngineView.java
+++ b/Modules/@babylonjs/react-native/android/src/main/java/com/babylonreactnative/EngineView.java
@@ -125,7 +125,7 @@ public final class EngineView extends FrameLayout implements SurfaceHolder.Callb
             String encoded = "";
             if (copyResult == PixelCopy.SUCCESS) {
                 ByteArrayOutputStream byteArrayStream = new ByteArrayOutputStream();
-                bitmap.compress(Bitmap.CompressFormat.JPEG, 100, byteArrayStream);
+                bitmap.compress(Bitmap.CompressFormat.JPEG, 80, byteArrayStream);
                 byte[] byteArray = byteArrayStream.toByteArray();
                 bitmap.recycle();
                 encoded = Base64.encodeToString(byteArray, Base64.DEFAULT);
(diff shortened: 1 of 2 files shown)


Modules/@babylonjs/react-native/android/src/main/java/com/babylonreactnative/EngineView.java
```diff
@@ -125,7 +125,7 @@ public void takeSnapshot() {
             String encoded = "";
             if (copyResult == PixelCopy.SUCCESS) {
                 ByteArrayOutputStream byteArrayStream = new ByteArrayOutputStream();
-                bitmap.compress(Bitmap.CompressFormat.JPEG, 100, byteArrayStream);
+                bitmap.compress(Bitmap.CompressFormat.JPEG, 80, byteArrayStream);
```
Alex-MSFT, line 128: Just a thought, maybe it makes sense to expose this as an argument so that consumers can set whatever desired compression quality they like.

ryantrem (approved):

Alex-MSFT (approved):

Modules/@babylonjs/react-native/android/src/main/java/com/babylonreactnative/EngineView.java
rarattay: We could do that, though I'm not sure its necessary with this change because there should not be any visual impact to the screenshots. This would also introduce a breaking change to anyone using screenshot, so I would recommend against this unless you feel strongly about it.

Alex-MSFT: I don't feel strongly either way.  I think eventually we want to replace this in our uses with the native capture apis in Babylon Native, so I think this is fine as is.

PR #251, opened 2021-06-17
Consume canvas polyfill to enable text support

chrisfromwork:

1. The babylonnative submodule reference is updated to contain changes for the canvas polyfill
2. The shared babylonnative initialization logic now initializes the canvas polyfill
3. babylon.js dependencies are updated to the versions supported by babylonnative

Note: we should wait and consume the following change in the babylonnative submodule update https://github.com/BabylonJS/BabylonNative/pull/805

We may also want to wait to update babylon.js dependencies to pick up the following change https://github.com/BabylonJS/Babylon.js/pull/10518 otherwise text still will not work

### DIFF
--- a/Apps/Playground/ios/Podfile.lock
+++ b/Apps/Playground/ios/Podfile.lock
@@ -467,9 +467,9 @@ EXTERNAL SOURCES:
 SPEC CHECKSUMS:
   boost-for-react-native: 39c7adb57c4e60d6c5479dd8623128eb5b3f0f2c
   CocoaAsyncSocket: 065fd1e645c7abab64f7a6a2007a48038fdc6a99
-  DoubleConversion: cde416483dac037923206447da6e1454df403714
+  DoubleConversion: cf9b38bf0b2d048436d9a82ad2abe1404f11e7de
   FBLazyVector: 7b423f9e248eae65987838148c36eec1dbfe0b53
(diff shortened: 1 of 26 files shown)


Apps/Playground/ios/Podfile.lock
```diff
@@ -492,7 +492,7 @@ SPEC CHECKSUMS:
   React-jsi: 5de8204706bd872b78ea646aee5d2561ca1214b6
   React-jsiexecutor: 124e8f99992490d0d13e0649d950d3e1aae06fe9
   React-jsinspector: 500a59626037be5b3b3d89c5151bc3baa9abf1a9
-  react-native-babylon: 67b39de846123d0846b28ce0e38671c7e3c4e04f
+  react-native-babylon: 858ff09e82f9c3de4f17d66e90d6ca08dd480bb4
```
chrisfromwork, line 495: @ryantrem is there more context around what this guid should map to?

ryantrem: I think it is a hash. I just ran `pod install` in the iOS directory. I'm not totally sure how the hash is generated for local packages.

ryantrem (approved):
Playground was updated to BJS 28, but the rest of the refs are still at 27. Will you please fix those up before merging?

rgerd (approved):

bghgary (approved):

PR #260, opened 2021-07-20
Support interacting with ARKit XR system in external native modules

chrisfromwork:
- [x] This change will update BabylonNative dependencies to pick up the following changes: https://github.com/BabylonJS/BabylonNative/pull/833

1. XrAnchorHelper and XrContextHelper are updated to support ARKit types
2. BabylonNative dependencies are updated

Testing changes:
1. Azures spatial anchors scenarios in external submodules utilize both of these headers. Changes were validated on both iOS and android.


Modules/@babylonjs/react-native/shared/XrAnchorHelper.h
```diff
@@ -183,3 +183,47 @@ namespace Babylon::Plugins::NativeXr
 }
 #endif
 #endif
+
+#if __has_include("IXrContextARKit.h")
+#include "IXrContextARKit.h"
+#if __has_include("jsi/jsi.h")
+namespace Babylon::Plugins::NativeXr
+{
+    bool TryGetNativeAnchor(facebook::jsi::Runtime& jsiRuntime, facebook::jsi::Value& jsAnchor, ARAnchor*& nativeAnchor)
```
rgerd, line 192: I think I prefer to use reference to pointer, but I feel like the pattern in OpenXR/Vulkan/D3D is usually to do pointer to pointer. Idk which is actually better, just wanted to _point_ it out

rgerd: Also, without looking too hard I think the first two references could be const all the way down

Modules/@babylonjs/react-native/shared/XrAnchorHelper.h
```diff
@@ -183,3 +183,47 @@ namespace Babylon::Plugins::NativeXr
...
+
+#if __has_include("IXrContextARKit.h")
+#include "IXrContextARKit.h"
+#if __has_include("jsi/jsi.h")
+namespace Babylon::Plugins::NativeXr
+{
+    bool TryGetNativeAnchor(facebook::jsi::Runtime& jsiRuntime, facebook::jsi::Value& jsAnchor, ARAnchor*& nativeAnchor)
+    {
+        nativeAnchor = nullptr;
+        uintptr_t nativeAnchorPtr{reinterpret_cast<uintptr_t>(nullptr)};
```
rgerd, line 195: nit: could use auto here

Modules/@babylonjs/react-native/shared/XrAnchorHelper.h
```diff
@@ -183,3 +183,47 @@ namespace Babylon::Plugins::NativeXr
 }
 #endif
 #endif
+
+#if __has_include("IXrContextARKit.h")
+#include "IXrContextARKit.h"
+#if __has_include("jsi/jsi.h")
```
rgerd, line 189: This feels like something that should be handled at the CMake level (like we do with AppRuntime* etc) instead of doing compile-time if's

rgerd (approved):

Modules/@babylonjs/react-native/shared/XrAnchorHelper.h
chrisfromwork: these files can be referenced by anyone independent of whether they use cmake

Modules/@babylonjs/react-native/shared/XrAnchorHelper.h
chrisfromwork: I agree on switching things that aren't getting set to constants. I think that a reference to a pointer is more legible than a double pointer personally.

chrisfromwork: I am going to leave things as they are for now. jsiRuntime.global() isn't marked as a const operation so we can get away with handing around a const jsiRuntime here.

Modules/@babylonjs/react-native/shared/XrAnchorHelper.h
chrisfromwork: going to leave as uintptr_t to stay consistent with rest of file/avoid editing the rest of the file.

PR #263, opened 2021-07-23
Add NativeOptimizations plugin

ryantrem:
**Describe the change**

This change adds the NativeOptimizations plugin to Babylon React Native. For now, this is primarily to improve MergeMeshes perf (about 9x).

**Screenshots**

*Before*:
MergeMeshesOptimizationBefore

*After*:
MergeMeshesOptimizationAfter

**Documentation**

N/A

**Testing**

I did all the tests for iOS and Android. I'm not setup to test Windows (it does compile though!).

### DIFF
--- a/Apps/Playground/package.json
+++ b/Apps/Playground/package.json
@@ -14,8 +14,8 @@
     "iosCmake": "node scripts/tools.js iosCMake"
   },
   "dependencies": {
-    "@babylonjs/core": "^5.0.0-alpha.30",
-    "@babylonjs/loaders": "^5.0.0-alpha.30",
+    "@babylonjs/core": "^5.0.0-alpha.34",
+    "@babylonjs/loaders": "^5.0.0-alpha.34",
     "@babylonjs/react-native": "file:../../Modules/@babylonjs/react-native",
     "@babylonjs/react-native-windows": "file:../../Modules/@babylonjs/react-native-windows",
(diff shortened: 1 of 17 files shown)


Apps/Playground/package.json
```diff
@@ -14,8 +14,8 @@
     "iosCmake": "node scripts/tools.js iosCMake"
   },
   "dependencies": {
-    "@babylonjs/core": "^5.0.0-alpha.30",
-    "@babylonjs/loaders": "^5.0.0-alpha.30",
+    "@babylonjs/core": "^5.0.0-alpha.34",
```
chrisfromwork, line 17: BabylonNative needs to be updated to this package before we update here.

chrisfromwork: not sure why this wasn't already done: https://github.com/BabylonJS/BabylonNative/blob/master/Apps/package.json

chrisfromwork (changes requested):
Please fix BJS dependencies in BN before updating BJS dependencies in BRN.

Apps/Playground/package.json
ryantrem: I didn't update BabylonNative because it's not a required change (and arguably didn't need to update it for the BRN package itself, just the test app), but I don't mind updating it in BN as well.

chrisfromwork (approved):
Feel free to check this in after updating BN commit

Issue #264, opened 2021-07-26
App crashes with @babylonjs/react-native above alpha.24 on Android


star-collector:
**Describe the bug**
it seems like something broke in alpha.25 and above. Any attempt to render 3D scene on Android leads to crash. Tested on several devices. iOS build works fine.

**To Reproduce**
1. Open app with @babylonjs/react-native alpha.25  on Android (example - https://github.com/star-collector/babylon-playground, I updated packages specifically for this issue, alpha.24 works)
2. App crashes

**Expected behavior**
App renders 3D scene without crash

**Other**
 - Platform: Android

ryantrem:
@star-collector I've tested alpha.25+ packages with the [PackageTest](https://github.com/BabylonJS/BabylonReactNative/tree/master/Apps/PackageTest/0.63.1) app without any issues on Android (haven't tried your app). Do you see exceptions in the logcat window?

star-collector:
> @star-collector I've tested alpha.25+ packages with the [PackageTest](https://github.com/BabylonJS/BabylonReactNative/tree/master/Apps/PackageTest/0.63.1) app without any issues on Android (haven't tried your app). Do you see exceptions in the logcat window?

Yes, I do.

<img width="1164" alt="Screenshot 2021-07-26 at 18 59 47" src="https://user-images.githubusercontent.com/19988178/127021122-ca129e50-3379-425a-bbe7-0f0f8591b54e.png">

And then some other logs later:

`WindowManager: WINDOW DIED Window{4908cce u0 com.playground/com.playground.MainActivity}`

ryantrem:
Oh interesting, I think that must be from this change (which happened between alpha.24 and alpha.25): https://github.com/BabylonJS/BabylonReactNative/commit/b4fec8c027c91ff68ca8006b3fd4484803f90e27

But I'm not sure why this isn't happening consistently on all devices/apps. @CedricGuillemet any ideas by chance?

@star-collector to help diagnose, can you try manually adding back the line (`implementation 'com.google.ar:core:1.14.0'`) to that file (`node_modules/@babylonjs/react-native/Android/build.gradle`) for your app and see if it fixes the problem?

star-collector:
@ryantrem Yes, it fixed the problem. My app rendered 3D without crash.

ryantrem:
Actually I forgot that the `PackageTest` app doesn't do any XR/MR/AR, so maybe we just broke it. I will do some more testing today.

[...]

Issue #218, opened 2021-05-08
Adding BabylonReactNative to a project already containing packages that require ARCore may fail to build


dmegue:
Adding BabylonReactNative to a project containing Viro React Community(https://github.com/ViroCommunity/viro) fails to build:

>[:arcore_client] /project/node_modules/react-viro/android/arcore_client/build/.transforms/8a1c382d15b6830d95b4fffc34498914/jetified-core-1.10.0/AndroidManifest.xml Warning:
Package name ‘com.google.ar.core’ used in: :arcore_client, com.google.ar:core:1.14.0.

>See http://g.co/androidstudio/manifest-merger for more information about the manifest merger.

>Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use ‘–warning-mode all’ to show the individual deprecation warnings.
See Command-Line Interface
265 actionable tasks: 5 executed, 260 up-to-date
/project/android/app/src/debug/AndroidManifest.xml:35:13-38 Error:
Attribute meta-data#com.google.ar.core.min_apk_version@value value=(190519000) from [:arcore_client] AndroidManifest.xml:35:13-38
is also present at [com.google.ar:core:1.14.0] AndroidManifest.xml:35:13-38 value=(191106000).
Suggestion: add ‘tools:replace=“android:value”’ to element at AndroidManifest.xml:33:9-35:41 to override.

>FAILURE: Build failed with an exception.

>What went wrong:
Execution failed for task ‘:app:processDebugMainManifest’.
Manifest merger failed with multiple errors, see logs`

Both require ARCore 1.14.0, but Viro React targets Android SDK 24 whilst BabylonReactNative targets SDK 28.

CedricGuillemet:
Hi @dmegue Are you referring to this particular line to remove ?
https://github.com/BabylonJS/BabylonReactNative/blob/5463b3a5605eb10ade59ade2eeab42cd65442820/Modules/%40babylonjs/react-native/android/build.gradle#L128

ryantrem:
I think we still need that one because that is the project that ultimately compiles all the Babylon Native code for Android. I think the one we want to remove is this one: https://github.com/BabylonJS/BabylonReactNative/blob/master/Package/Android/build.gradle#L80. That's the one @dmegue tested removing in their project.

A bit more context here: https://forum.babylonjs.com/t/issue-setting-up-babylonjs-on-rn/20647/5?u=ryantrem

dmegue:
[...]

PR #265, opened 2021-07-26
Add back ARCore

ryantrem:
**Describe the change**

This change adds back ARCore to the packages build.gradle file for Android. This *unfixes* https://github.com/BabylonJS/BabylonReactNative/issues/218 (we'll need to come up with a different solution), but fixes https://github.com/BabylonJS/BabylonReactNative/issues/264.

I'm not sure how this worked when I tested it before, because now when I test the current package in PackageTest (which doesn't even use XR) it crashes on launch with the missing so exception mentioned in #264. I did do a clean build and cleaned my npm cache, so maybe not having done so previously was somehow preventing this error when I previously tested, not sure. :/

**Screenshots**

N/A

**Documentation**

N/A

**Testing**

Tested on Android (the only platform affected by this change).

### DIFF
--- a/Package/Android/build.gradle
+++ b/Package/Android/build.gradle
@@ -77,6 +77,7 @@ repositories {
 dependencies {
     //noinspection GradleDynamicVersion
     implementation 'com.facebook.react:react-native:+'  // From node_modules
+    implementation 'com.google.ar:core:1.14.0'
 }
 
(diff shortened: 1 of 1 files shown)


CedricGuillemet:
This is weird. I've tested on a freshly cloned repo.

CedricGuillemet (approved):

PR #280, opened 2021-08-24
Implement FontFace polyfill to expose loading of custom fonts

Alex-MSFT:
**Describe the change**

This PR implements a temporary FontFace polyfill to allow for a TypeSafe way to load custom fonts via either an ArrayBuffer or URL

**Screenshots**

**Documentation**

N/A

**Testing**

Tested on iOS and Android


Modules/@babylonjs/react-native/FontFace.ts
```diff
@@ -0,0 +1,35 @@
...
+  private status: "unloaded" | "loading" | "loaded" | "error" = "unloaded";
+  public constructor(public readonly family: string, source: string | ArrayBuffer) {
+    this.source = source;
+  }
+
+  public async load(): Promise<void> {
+    try {
+      this.status = "loading";
+      if (typeof this.source === 'string') {
+        this.source = await Tools.LoadFileAsync(this.source as string);
```
ryantrem, line 20: The cast shouldn't be needed as TypeScript will have narrowed it to `string` after the conditional check on line 19.
```suggestion
        this.source = await Tools.LoadFileAsync(this.source);
```

Modules/@babylonjs/react-native/FontFace.ts
```diff
@@ -0,0 +1,35 @@
+import { Tools } from '@babylonjs/core';
+
+// Declare _native to get access to NativeCanvas.
+declare var _native: any;
+
+/**
+ * Partial Polyfill for FontFace Web API to wrap the NativeCanvas object.
+ */
+export class FontFace {
+  private source: string | ArrayBuffer | undefined;
```
ryantrem, line 10: Maybe just `string` for now if that's all we are going to support?

ryantrem (approved):

ryantrem:
Related discussion: https://github.com/BabylonJS/BabylonNative/pull/441#discussion_r652157811

Modules/@babylonjs/react-native/FontFace.ts
Alex-MSFT: I'm assigning the source to undefined after we load the font so that we don't hold two copies of it in memory

PR #282, opened 2021-09-10
Update to Babylon 5.0.0-alpha.44

bghgary:
This change updates Babylon to 5.0.0-alpha.44 to address issues with XR clearing. It also removes the workaround for removing the extra clear that should be no longer necessary.

See https://github.com/BabylonJS/BabylonNative/issues/871 for more context.

Additional changes:
- Changed PackageTest to point to public released npm packages.
- Updated NativeCapture to match changes to Babylon.js framebuffers.


ryantrem:
Shouldn't package-lock.json be updated for the PackageTest apps?

PR #287, opened 2021-09-30
Fix missing Swift Libraries error when building via XCode 13

Alex-MSFT:
**Describe the change**

When building react-native applications via Xcode 13, certain Swift library paths are no longer included by default.  This PR corrects this by manually adding them to the list of library search paths to the pbxproj files.  You can see the equivalent of this change for the react-native template app here: https://github.com/facebook/react-native/commit/eb938863063f5535735af2be4e706f70647e5b90

**Screenshots**
N/A

**Documentation**

N/A

**Testing**

Built the Playground and package test apps in both release and debug mode.


Apps/Playground/ios/Playground.xcodeproj/project.pbxproj
```diff
@@ -634,6 +634,46 @@
 				ENABLE_BITCODE = NO;
 				INFOPLIST_FILE = Playground/Info.plist;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
+				LIBRARY_SEARCH_PATHS = (
```
ryantrem, line 637: What's all this? Something XCode 13 did on its own?

ryantrem (approved):

Apps/Playground/ios/Playground.xcodeproj/project.pbxproj
Alex-MSFT: Yes I believe this is because before this field was auto-populated by app template. Now that we have added a custom path everything gets dumped to the pbxproj file.

PR #297, opened 2021-10-20
Update to latest BabylonNative with 5.0.0-alpha.57

bghgary:
Tested with Android

This change includes the performance improvements with using command buffers. It also locks the babylonjs npm packages to a specific version.


bghgary:
@star-collector @Alex-MSFT tried to repro but wasn't able to. Is your scenario similar to the repo you sent before? Can you help us repro this problem?

star-collector:
> @star-collector We are going to merge this. Can you create a new issue so we can continue tracking?

Ok

Issue #152, opened 2021-02-05
Transparent background not working - scene.clearColor ignores alpha channel


swese44:
When setting `scene.clearColor`, `Color4`'s alpha channel is ignored. The background color is always the RGB values.

Basic example:

```ts
import { useEngine, EngineView } from "@babylonjs/react-native"
import {
  Scene,
  Camera,
  MeshBuilder,
  FreeCamera,
  Vector3,
  HemisphericLight,
  Color4,
} from "@babylonjs/core"
import React, { FC, useEffect, useState } from "react"

const engineStyle = { flex: 1, backgroundColor: "transparent" }

export const BabylonTest: FC = () => {
  const engine = useEngine()
  const [camera, setCamera] = useState<Camera>()

  const createScene = function () {
    const scene = new Scene(engine)
    scene.clearColor = new Color4(0, 0, 0, 0)
    const camera = new FreeCamera("camera1", new Vector3(0, 5, -10), scene)
    setCamera(camera)
    camera.setTarget(Vector3.Zero())
    const light = new HemisphericLight("light", new Vector3(0, 1, 0), scene)
    light.intensity = 0.7
    const sphere = MeshBuilder.CreateSphere("sphere", { diameter: 2, segments: 32 }, scene)
    sphere.position.y = 1
    const ground = MeshBuilder.CreateGround("ground", { width: 6, height: 6 }, scene)

    return scene
  }

  useEffect(() => {
    if (engine) {
      const scene = createScene()
    }
  }, [engine])

  return <EngineView style={engineStyle} camera={camera} />
}
```

<img width="374" alt="Screen Shot 2021-02-04 at 11 54 18 PM" src="https://user-images.githubusercontent.com/471731/107005548-b3635280-6744-11eb-953f-dacc6e7bbbbb.png">

ryantrem:
I've confirmed that setting `[super setOpaque:false];` in `EngineView`'s `init` fixes this on iOS.

On Android, in `EngineView`'s constructor I tried this:
```java
        this.setZOrderOnTop(true);
        this.getHolder().setFormat(PixelFormat.TRANSPARENT);
```
Which kind of works, but it makes it so that UI cannot be drawn on top of `EngineView`, which obvious breaks other scenarios. Needs a bit more investigation...

ryantrem:
[...]

PR #314, opened 2021-11-18
Feat/transparent background

daenash:
**Describe the change**

Issue #152

The `EngineView` is extended with an `isTransparent` boolean flag which is not required to set, on default it's false.

If it's set to `true` this happens:

- **Android side**: The SurfaceView is changed to a `TextureView` and the root `FrameLayout` is extended to implement a `TextureView.SurfaceTextureListener`. The listener itself is using a similar runnable implementation as the SurfaceView. Note: There is no XR related implementation.
- **iOS side**: The `MKTView` is extended with an isTransparent `@property` and it sets whether the opaque flag is true or false.

**Screenshots**

_Nope, sorry 🦄_

**Documentation**

[Updated](Modules/@babylonjs/react-native/README.md)

**Testing**

I need help testing on Windows
For the other platforms I tested it and it worked, so YES


ryantrem:
Thanks for this contribution! There are two main questions for me:
1. Does the Android solution really work? It seems like the transparency flag needs to be passed to the `EngineView` instance after it is created.
2. Can we still enable the XR on Android even when transparency is enabled on the primary view? I don't see any technical limitations, and it seems like this behavior would match what iOS is doing.

ryantrem:
@chrisfromwork, @rgerd - do you know what needs to happen to at least stub this out on Windows (e.g. just making sure this change doesn't break on Windows even if you don't specify the `isTransparent` property)?

Modules/@babylonjs/react-native/android/src/main/java/com/babylonreactnative/EngineViewManager.java
```diff
@@ -1,29 +1,39 @@
...
 
     @NonNull
     @Override
     public String getName() {
         return "EngineView";
     }
 
+    @ReactProp(name = "isTransparent")
+    public void setIsTransparent(EngineView view, Boolean _isTransparent) {
+      isTransparent = _isTransparent;
```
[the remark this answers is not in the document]
daenash: Well, strangely it happened for me before the initialization because it was working correctly. But yeah debugging it step by step might show otherwise.

daenash:
@ryantrem 

Thank you for the review! 🙌  Well basically I'm not a native developer so sorry for the flaws I might have made here 🙃 I'll try to do some changes based on your suggestions, but I don't really know when. 

For my project I just patched the package with the transparent solutions, so there is no chance for it being not transparent. This was the case I needed.

I'm not quite understand what that XR view is, and for my purpose I haven't needed it. I'll need to take a closer look at that.

ryantrem:
@daenash I think your code is really close, mainly just need to sort out how the transparency flag is passed to the Android `EngineView`! Really appreciate the contribution!

The XR view is the augmented reality view when you use Babylon.js to create and start a WebXR session (https://doc.babylonjs.com/divingDeeper/webXR/webXRSessionManagers). So the `EngineView` classes have a different view for regular rendering vs. XR rendering because they just work really differently. But, the XR view should be supported (and work the same way) regardless of whether transparency is enabled. From what I can tell, that's actually what your implementation does for iOS since you explicitly ignore the transparency flag on the xr view.

daenash:
@ryantrem

With more code inspection I realized maybe it would be good the separate the two engine views to two separate components

One for transparent background one opaque like this

```tsx
<EngineView />
<EngineViewTransparent />
```

And in the Android part we would implement two different classes which are extending a JAVA abstract class which contains the Babylon updating functionalities. So this way these classes would only need to implement the `TextureView.SurfaceTextureListener` and `SurfaceHolder.Callback` separately and extend a so called `EngineViewBase` class.

For iOS as only a flag would change on the View some other approach would be nice. I'll think about it.
Until that in Typescript maybe we could only render an EngineView with a transparency flag for iOS.

I'll try to create a structure for what I mean.

ryantrem:
That's an interesting idea. I think I'd be more inclined to keep it as one control though to keep the API simple. If we add more options like this, we could end up with an explosion of control variants, which wouldn't scale.

CedricGuillemet:
Hi @daenash , any update on this PR?

daenash:
@CedricGuillemet Hi! No, unfortunately I had no time to follow up with this. 😿  Feel free to continue working on it if you feel so.

CedricGuillemet:
I changed the Android impl:
- always using a `TextureView` so opacity can easily be changed at runtime
- removed transparency boolean in engine view manager
- added code in mandatory callbacks when view size changed

ryantrem:
> I changed the Android impl:
> 
> * always using a `TextureView` so opacity can easily be changed at runtime
> * removed transparency boolean in engine view manager
> * added code in mandatory callbacks when view size changed

I don't think we should switch to `TextureView` unconditionally as it has perf implications. My understanding is that with `TextureView`, since the compositor draws it, frame rate is limited to the compositor frame rate, and there is an extra layer of synchronization (e.g. Babylon's frame timing compared to the Android UI compositor's frame timing). I think it would be much safer to only use `TextureView` when transparency is needed. There is some discussion about this in the associated issue as well.

Modules/@babylonjs/react-native/ios/EngineViewManager.mm
```diff
@@ -22,14 +24,25 @@ @implementation EngineView {
...
         super.depthStencilPixelFormat = MTLPixelFormatDepth32Float;
     }
     return self;
 }
 
+- (void)setIsTransparentFlag:(NSNumber*)isTransparentFlag {
+    BOOL isTransparent = [isTransparentFlag intValue] == 1;
+    if(isTransparent){
+        [self setOpaque:NO];
+        [self.window setOpaque:NO];
```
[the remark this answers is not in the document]
CedricGuillemet: No need to set opacity on window. I removed it.

Modules/@babylonjs/react-native/android/src/main/java/com/babylonreactnative/EngineView.java
```diff
@@ -114,13 +168,20 @@ public void takeSnapshot() {
...
         }
+        PixelCopy.request(sourceSurface, bitmap, getOnPixelCopyFinishedListener(bitmap, helperThread), helperThreadHandler);
+    }
 
-        // Request the pixel copy.
-        PixelCopy.request(surfaceView, bitmap, (copyResult) ->  {
+    // ---------------------------------------------------------------------------------------------
+    // Returns the listener for the PixelCopy.request function call
+    @NonNull
+    private PixelCopy.OnPixelCopyFinishedListener getOnPixelCopyFinishedListener(Bitmap bitmap, HandlerThread helperThread) {
```
ryantrem, line 183: Do we really need a function that returns a function for this? Why not just leave it inline as it originally was?

Modules/@babylonjs/react-native/android/src/main/java/com/babylonreactnative/EngineView.java
```diff
@@ -89,6 +114,35 @@ public void surfaceDestroyed(SurfaceHolder surfaceHolder) {
         this.renderRunnable = null;
     }
 
+    @Override
+    public void onSurfaceTextureAvailable(@NonNull SurfaceTexture surfaceTexture, int i, int i1) {
+        this.startRenderLoop();
+        transparentSurface = new Surface(surfaceTexture);
```
ryantrem, line 120: `this.transparentSurface` instead of just `transparentSurface` for all the references to this member variable.

Modules/@babylonjs/react-native/ios/EngineViewManager.mm
```diff
@@ -55,6 +66,10 @@ - (void)drawRect:(CGRect)rect {
     if ([BabylonNativeInterop isXRActive]) {
         if (!xrView) {
             xrView = [[MTKView alloc] initWithFrame:self.bounds device:self.device];
+            if(self.isTransparent == YES){
+                [xrView setOpaque:NO];
+                [xrView.window setOpaque:NO];
+            }
```
ryantrem, line 72: This should not be needed, right? Any transparency settings should never affect the xr view, right?

Modules/@babylonjs/react-native/android/src/main/java/com/babylonreactnative/EngineViewManager.java
```diff
@@ -20,6 +23,11 @@ public String getName() {
         return "EngineView";
     }
 
+    @ReactProp(name = "isTransparent")
+    public void setIsTransparent(EngineView view, Boolean isTransparent) {
```
ryantrem, line 27: Does this get called for the initial prop value? I would expect it to, and if so it seems like the constructor doesn't need to create a default SurfaceView/TextureView.

ryantrem (approved):

Modules/@babylonjs/react-native/android/src/main/java/com/babylonreactnative/EngineViewManager.java
CedricGuillemet: if no transparency property is set at engineView creation, this method is not called. So, the constructor needs to build the view with default parameters (transparency = false).

Modules/@babylonjs/react-native/ios/EngineViewManager.mm
CedricGuillemet: Removed and tested.

Modules/@babylonjs/react-native/android/src/main/java/com/babylonreactnative/EngineView.java
CedricGuillemet: I find it more clear that way.

Modules/@babylonjs/react-native/android/src/main/java/com/babylonreactnative/EngineView.java
```diff
@@ -89,6 +114,35 @@ public void surfaceDestroyed(SurfaceHolder surfaceHolder) {
...
+        BabylonNativeInterop.updateView(transparentSurface);
+    }
+
+    @Override
+    public void onSurfaceTextureSizeChanged(@NonNull SurfaceTexture surfaceTexture, int i, int i1) {
+        transparentSurface = new Surface(surfaceTexture);
```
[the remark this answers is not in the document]
CedricGuillemet: Indeed. I've added a new method to release and acquire the new one.

PR #323, opened 2022-01-13
React Native 0.65 Android/iOS PR 2/n

CedricGuillemet:
This PR is independant of https://github.com/BabylonJS/BabylonReactNative/pull/322 but will not pass CI validation until 322 is in.

The project has been generated on MacOS with this command line :
```
npx react-native init Playground --version 0.65.0 --template react-native-template-typescript
```
Once done, I've updated package.json with babylon dependencies and copy/pasted missing devdepencies from 0.64.
I've also done some more steps (xcode workspace, manifest, ... ) that are listed in the doc.

To test it, clone it, make a symlink :
```
cd Apps/Playground
npm install
npm run select 0.64
cd Playground
npm run ios OR npm run android
```

- [x] Android
- [x] iOS (device + simulator)

AR tested on both platforms


Apps/Playground/0.65/android/app/proguard-rules.pro
```diff
@@ -0,0 +1,10 @@
+# Add project specific ProGuard rules here.
+# By default, the flags in this file are appended to flags specified
+# in /usr/local/Cellar/android-sdk/24.3.3/tools/proguard/proguard-android.txt
+# You can edit the include path and order by changing the proguardFiles
+# directive in build.gradle.
+#
+# For more details, see
+#   http://developer.android.com/guide/developing/tools/proguard.html
+
+# Add any project specific keep options here:
```
ryantrem, line 10: A couple lines are needed here to make Hermes work. Probably worth copying those over from the 0.64 project.

ryantrem:
Seems like the package-lock.json file is missing. If you haven't already, you should delete the yarn.lock file, then do an `npm install` (with npm 8.1) to generate the package-lock.json and add that file to this PR.

ryantrem (approved):

Apps/Playground/0.65/android/app/proguard-rules.pro
CedricGuillemet: Done with documentation update as well.

PR #329, opened 2022-01-24
Windows transparency 

CedricGuillemet:
Same as https://github.com/BabylonJS/BabylonReactNative/pull/314 but for Windows.

Property is forwarded to the view.

to test, add `isTransparent` true in App.tsx:
```
<EngineView camera={camera} onInitialized={onInitialized} displayFrameRate={true} isTransparent={true}/>
```

and set clear color with an alpha=0:

```
scene.clearColor = new Color4(0, 1, 0, 0);
```


Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
```diff
@@ -168,4 +170,22 @@ namespace winrt::BabylonReactNative::implementation {
...
+
+        for (auto const& pair : propertyMap)
+        {
+            auto const& propertyName = pair.first;
+            auto const& propertyValue = pair.second;
+
+            if (propertyName == "isTransparent")
+            {
+                bool isTransparent = propertyValue.AsBoolean();
+                auto engineView = static_cast<winrt::Windows::UI::Xaml::Controls::SwapChainPanel>(*this);
```
ryantrem, line 186: Why is this cast needed?

tymlipari: This is also unsafe with C++/WinRT. If you need to cast between types always use the `as` or `try_as` functions.

But since `EngineView` derives from `SwapChainPanel` you should be able to just directly call `CompositeMode`. IntelliSense might complain because of C++/WinRT's extreme usage of CRTP, but I think it should still compile.

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
```diff
@@ -168,4 +170,22 @@ namespace winrt::BabylonReactNative::implementation {
...
+    void EngineView::UpdateProperties(IJSValueReader const& reader)
+    {
+        auto const& propertyMap = JSValueObject::ReadFrom(reader);
+
+        for (auto const& pair : propertyMap)
+        {
+            auto const& propertyName = pair.first;
+            auto const& propertyValue = pair.second;
+
+            if (propertyName == "isTransparent")
```
tymlipari, line 183: I admit I'm not 100% sure how the native properties propagate in React Native, but is there a possibility of clearing `isTransparent` from the map at runtime? Do we need to also handle the case where `isTransparent` goes from being true to just unset? Right now that'd currently result in the view remaining transparent.

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
tymlipari, line 186: Nit: we already have `using namespace` directives for all of the `winrt::*` namespaces we use. You don't need to fully qualify these types.

tymlipari (approved):
Generally looks fine, left a few nit comments

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
CedricGuillemet: Fixed.

CedricGuillemet: cast removed.

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
CedricGuillemet: Same as iOS/Android version, transparency doesn't change until property is changed.

PR #337, opened 2022-01-31
React Native version switch for publishing

CedricGuillemet:
This PR allows switching React Native version based on published release version.
When publishing a new release, the Major and Minor numbers are used to choose the React Native version.
For example:
A release version `0.64.alpha.39` will trigger a build and publish with React Native 0.64 whereas `0.65.New` will use React Native 0.65.

In case of a typo or a version mismatch, the command `npm run select` will fail the build.

Tested without final npm publish here : https://github.com/CedricGuillemet/BabylonReactNative-1/actions/runs/1772840138


.github/workflows/publish.yml
```diff
@@ -3,12 +3,28 @@ on:
...
+        with:
+          text: ${{ github.event.release.name }}
+          regex: '^(\d+\.)?(\d+)'
+      - name: Release infos
+        run: |
+          echo "- Tag ${GITHUB_REF/refs\/tags\//}"
+          echo "- Name: ${{ github.event.release.name }}"
+          echo "- Description: ${{ github.event.release.body }}"
+      - id: stepVersion
+        run: echo "::set-output name=version::${{ steps.regex-match.outputs.match }}"
```
ryantrem, line 24: I don't follow what's happening on this line. When I saw the `echo` I thought it was just logging or something, but it looks very syntaxy so I'm guessing it's doing something else?

ryantrem: Oh I think I see, this is some syntax for setting the `select-react-native-version` job output, which is accessed later via `needs.select-react-native-version.outputs.version`, correct?

ryantrem:
Not really specific to this PR, but from all the PRs related to this, do we patch the `package.json` to correctly set the React Native dependency? E.g. the BRN `0.64.alpha-39` package should depend on RN `>=0.63.1 <0.65.0`, and the `0.65.alpha-39` package should depend on RN `>=0.65.0` (or something like that).

.github/workflows/publish.yml
CedricGuillemet: Yes, that's correct. The syntax for forwarding parameters is indeed special.

ryantrem:
> > Not really specific to this PR, but from all the PRs related to this, do we patch the `package.json` to correctly set the React Native dependency? E.g. the BRN `0.64.alpha-39` package should depend on RN `>=0.63.1 <0.65.0`, and the `0.65.alpha-39` package should depend on RN `>=0.65.0` (or something like that).
> 
> Which package.json are you referring to?

The one included in the BRN package, so that when someone consumes BRN 0.64-alpha.39 (for example) it has a dependency on RN 0.64 (and 0.65-alpha.39 for example has a dependency on RN 0.65).

CedricGuillemet:
Ho! I got it! I've added a new function to patch both package.json when publishing. Nothing should change by default.

ryantrem (approved):

.github/workflows/publish.yml
```diff
@@ -23,14 +39,14 @@ jobs:
       - name: NPM Install (Playground)
         run: npm install
         working-directory: ./Apps/Playground
-      - name: NPM Install (React Native ${{ env.BRN_Version }})
-        run: npm run select --reactNative ${{ env.BRN_Version }}
+      - name: NPM Install (React Native ${{ needs.select-react-native-version.outputs.version }})
```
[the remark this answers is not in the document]
CedricGuillemet: yes, that will be better.

PR #342, opened 2022-02-17
Try changing Windows to use older build pool with 10.0.18362 SDK

tymlipari:
**Describe the change**

Windows builds are currently failing due to the 10.0.18362.0 SDK not being available. A recent change to the Windows agents removed this SDK. This change tries to switch us to using an older build agent that still has the SDK available while we work on migrating the projects to use a newer version of the WinSDK.


ryantrem (approved):

tymlipari:
Filed #342 to track applying a better fix

PR #345, opened 2022-02-22
Add Engine Render Hooks

sebavan:
This enables the possibility to use engine without scenes or camera through render hooks without loosing the appState and life cycle automated management.


sebavan:
I am not fully certain of the local test procedure to follow ? so do not hesitate if there is anything else I should double check ?

Modules/@babylonjs/react-native/EngineHook.ts
```diff
@@ -185,6 +185,44 @@ declare var _native: {
     }
 }
 
+export function useAppState(): string {
```
ryantrem, line 188: If we are exporting these, I guess we are saying they are part of the BRN public contract? Or should we not consider them part of the public contract and not re-export them from `index.js`?

PR #355, opened 2022-03-08
deps: lock @babylonjs/core version #354

### LINKED ISSUE #354: protocol version mismatch 2 (Native) !== 4 (JS)
**For questions, please use the [forum](https://forum.babylonjs.com)**.

**Describe the bug**
I think this is not a bug, @babylonjs/react-native just hasn't yet followed the changes going on in @babylonjs/core. [This](140ce5e6a36a0315dc67236bbf0e7238e2ab8c4b) latest commit that this package works with,  (release of @babylonjs/core@v5.0.0-beta.11)

**To Reproduce**
1. Install latest rc from @babylonjs/core
2. Try to render anything

**Expected behavior**
@babylonjs/react-native works with latest @babylonjs/core

I don't intend this as a bug report, sorry if I chose the wrong forum to mark this, but people will find it here the best probably. I know that you'll eventually follow ProtocolVersion of @babylonjs/core. I just intend to mark this issue for people that update their @babylonjs/core version automatically and break their app, and don't know where to start. Just downgrade @babylonjs/core (and loaders) to v5.0.0-beta.11 and it will work with @babylonjs/react-native@0.64.0-alpha.47 until this package catches up.

Edit:

Probably this should be updated in the package.json of @babylonjs/react-native as peerDependency
```
"@babylonjs/core": "^5.0.0-alpha.65",
```
will break things for many people.

f4z3k4s:
**Describe the change**

Locks the last version of @bablonjs/core that works with this package. Fixes #354.

### DIFF
--- a/Modules/@babylonjs/react-native/package.json
+++ b/Modules/@babylonjs/react-native/package.json
@@ -28,7 +28,7 @@
     "semver": "^7.3.2"
   },
   "peerDependencies": {
-    "@babylonjs/core": "^5.0.0-alpha.65",
+    "@babylonjs/core": "5.0.0-beta.11",
     "react": ">=16.13.1",
     "react-native": ">=0.63.1",
     "react-native-permissions": "^2.1.4"
@@ -36,7 +36,7 @@
   "devDependencies": {
     "@babel/core": "^7.8.4",
     "@babel/runtime": "^7.8.4",
-    "@babylonjs/core": "5.0.0-alpha.65",
+    "@babylonjs/core": "5.0.0-beta.11",
     "@rnw-scripts/eslint-config": "0.1.6",
     "@rnw-scripts/ts-config": "0.1.0",
     "@types/base-64": "^0.1.3",


ryantrem:
The problem with this approach is that it locks Babylon React Native to a specific version of Babylon.js. If say a bug is fixed in BJS, then an app consuming BRN and BJS can't just update BJS to get the bug fix. They have to wait until BRN moves to that version of BJS. I think other ways we could address this are:
1. Use semantic versioning in BJS (unlikely)
2. Try to make protocol changes between BN/BJS more backward compatible (so newer versions of BJS work with older versions of BN)
3. Have better error messaging when there is a protocol mismatch (like if you are too far ahead in BJS, have a map of BJS versions to protocol versions so we can suggest an older BJS version to use)

f4z3k4s:
You're right. I just felt it's a bad experience that in this form, with having `^` before the version, the package just doesn't work currently. Being not able to have a bugfix is probably better, than having to figure out why the package you just added to your project doesn't work at all. You're right in the fact that this makes people rely on you. In my opinion, it's a good tradeoff as you probably know better which versions are compatible and why versions break, and you can update when you feel there are 2 versions that work together, and not put give this responsibility for the users of your package. I accept that we think differently and you won't merge this PR, it's totally okay. Thank you for your reply.

PR #361, opened 2022-03-18
New package system

CedricGuillemet:
- New package `react-native-iosandroid` in `Modules\@babylonjs` that contains ios and android folder. Deps on react-native.
- `react-native` package contains the Shared folder and compiled TS to .js. Dependency on @babylon/core but no deps on native
-  react-native-windows alsmost unchanged
- publish workflow now uses reusable workflows to build TS, iOS-Android and Windows
- react-native-iosandroid and react-native-windows have a peerdependecy on react-native with version of the package
- podspec moved to ios-android package

Most of the changes are in `gulpfile.js`.

You can check the result of publish Action on my repo: https://github.com/CedricGuillemet/BabylonReactNative-1/actions

Package testing:
Download artifacts. zip, unzip then `npm install path\to\unzipped\folder`


SergioRZMasson:
Tested generated artifacts on a new React Native project on Windows:

Android - react native 0,64 - is working as expected.
Android - react native 0.65 - is working as expected.

.github/workflows/typescript.yml
```diff
@@ -0,0 +1,29 @@
+name: 'build ios-android'
```
ryantrem, line 1: Should the name be 'Build TypeScript'?

.github/workflows/publish.yml
```diff
@@ -3,94 +3,32 @@ on:
...
-        working-directory: ./Package
-      - name: Upload Assembled Folder
-        uses: actions/upload-artifact@v2
-        with:
-          name: 'Assembled'
-          path: Package/Assembled
+  build-android-ios-064:
+    uses: ./.github/workflows/ios_android.yml
+    with:
+      react-native-version: 0.64
```
ryantrem, line 10: I guess we can't use the build mtarix stuff for this?

Modules/@babylonjs/react-native-iosandroid/.gitattributes
```diff
@@ -0,0 +1 @@
+*.pbxproj -text
```
ryantrem, line 1: Do we really need this?

Modules/@babylonjs/react-native-iosandroid/.gitignore
```diff
@@ -0,0 +1,51 @@
+# OSX
```
ryantrem, line 1: Shouldn't this file mostly be a move (from react-native)?

Modules/@babylonjs/react-native-iosandroid/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "license": "MIT",
+  "licenseFilename": "LICENSE",
+  "readmeFilename": "README.md",
+  "dependencies": {
+    "base-64": "^0.1.0",
+    "semver": "^7.3.2"
+  },
+  "peerDependencies": {
+    "@babylonjs/core": "^5.0.0-rc.4",
+    "@babylonjs/react-native": "version",
```
ryantrem, line 32: Doesn't having this be "version" mess up the local build? Should this be 0.0.1?

Modules/@babylonjs/react-native-iosandroid/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "homepage": "https://github.com/BabylonJS/BabylonReactNative#readme",
+  "license": "MIT",
+  "licenseFilename": "LICENSE",
+  "readmeFilename": "README.md",
+  "dependencies": {
+    "base-64": "^0.1.0",
+    "semver": "^7.3.2"
+  },
+  "peerDependencies": {
+    "@babylonjs/core": "^5.0.0-rc.4",
```
ryantrem, line 31: It seems like this dependency shouldn't be here. Instead it seems like this package should just depend on @babylonjs/react-native, and @babylonjs/react-native depends on @babylonjs/core in turn.

CedricGuillemet:
@ryantrem When testing the artifacts on iOS, I get an error with `-lBabylonNative` not found. I guess it's because the search path has not been updated (react-native => react-native-iosandroid). I spent some time trying to find where to do the change without success. Do you now where the search path must be updated?

Modules/@babylonjs/react-native-iosandroid/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "@babylonjs/react-native": "version",
+    "react": ">=16.13.1",
+    "react-native": ">=0.63.1",
+    "react-native-permissions": "^3.0.0"
```
ryantrem, line 35: Maybe similar comment here, about these being transitive dependencies through @babylonjs/react-native (I think that is the only place we actually have TypeScript dependencies on these). I think for react-native we still need the dependency at build time, but maybe that means react-native should be a devDependency for this module?

Modules/@babylonjs/react-native-iosandroid/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "devDependencies": {
+    "@babel/core": "^7.8.4",
+    "@babel/runtime": "^7.8.4",
+    "@babylonjs/core": "5.0.0-rc.4",
+    "@rnw-scripts/eslint-config": "0.1.6",
+    "@rnw-scripts/ts-config": "0.1.0",
+    "@types/base-64": "^0.1.3",
+    "@types/jest": "^25.2.1",
+    "@types/react": "^16.9.32",
+    "@types/react-native": "^0.63.1",
```
ryantrem, line 46: I think we can remove this since we shouldn't have any TypeScript in this module?

Modules/@babylonjs/react-native-windows/windows/CMakeLists.txt
```diff
@@ -5,8 +5,8 @@ project(ReactNativeBabylon)
 
 set(CMAKE_CXX_STANDARD 17)
 
-set(BABYLON_REACT_NATIVE_DIR "${CMAKE_CURRENT_LIST_DIR}/../../react-native")
-include(${BABYLON_REACT_NATIVE_DIR}/shared/CMakeLists.txt)
+set(BABYLON_REACT_NATIVE_DIR "${CMAKE_CURRENT_LIST_DIR}/../../react-native-iosandroid")
+include(${CMAKE_CURRENT_LIST_DIR}/../../react-native/shared/CMakeLists.txt)
```
ryantrem, line 9: Why this change?

Modules/@babylonjs/react-native/package.json
```diff
@@ -29,9 +29,7 @@
   },
   "peerDependencies": {
     "@babylonjs/core": "^5.0.0-rc.4",
-    "react": ">=16.13.1",
-    "react-native": ">=0.63.1",
-    "react-native-permissions": "^2.1.4"
```
ryantrem, line 34: Hmm so you were thinking move these dependencies to the binary packages, and I was thinking have these only in this package. I guess it's weird because we have both TypeScript and native dependencies on these packages. I think I feel like this is a lower level package that does in fact have these dependencies, so I still think it probably makes more sense to have them here, but if you have a different perspective I'm definitely open.

Package/gulpfile.js
```diff
@@ -543,35 +570,50 @@ const patchPackageVersion = async () => {
...
       packageJsonWindows.peerDependencies['react-native-windows'] = '>=0.63.1 <0.65.0';
     } else {
-      packageJson.peerDependencies['react-native'] = '>=0.65.0';
       packageJsonWindows.peerDependencies['react-native'] = '>=0.65.0';
+      packageJsoniOSAndroid.peerDependencies['react-native'] = '>=0.65.0';
       packageJsonWindows.peerDependencies['react-native-windows'] = '>=0.65.0';
     }
 
+    // release version
+    const releaseVersion = (process.argv[4] == '--releaseVersion') ? process.argv[5] : ((process.argv[5] == '--releaseVersion') ? process.argv[6] : '');
```
ryantrem, line 589: Does this mean when testing the package build locally you now have to pass in a version as an arg?

Package/gulpfile.js
```diff
@@ -19,11 +19,40 @@ function exec(command, workingDirectory = '.', logCommand = true) {
...
 }
 
+function checkDirectory(actualList, expectedList) {
+  const extras = actualList.filter(path => !expectedList.includes(path));
+  const missing = expectedList.filter(path => !actualList.includes(path));
+
+  let isValid = true;
+
+  if (extras.length !== 0) {
+    console.error(chalk.white.bgRedBright(`The Assembled directory contains unexpected files:`));
```
ryantrem, line 29: Maybe pass a directory name arg so we can print out the right directory here (and not just "Assembled")?

ryantrem:
Regarding this file: Modules/@babylonjs/react-native-iosandroid/package-lock.json
GitHub diff says: "22,442 additions, 0 deletions not shown because the diff is too large. Please use a local Git client to view these changes."
I guess this is because there is still a package-lock.json file in @babylonjs/react-native, but is it effectively just a file move?

ryantrem:
One other change I would include is to update the `VersionValidation.ts` file to check the `Platform.OS` and verify that you have the right binary package and version installed, and if not, emit an error. This will make it easy for people updating to the new @babylonjs/react-native package understand there are additional dependencies. It would probably also be good to have an exported variable from `VersionValidation.ts` to indicate whether the binary packages are present, and if not, have useEngine and EngineView no-op or something (and not explode).

.github/workflows/publish.yml
```diff
@@ -103,30 +41,45 @@ jobs:
...
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
-      - name: Version & Publish Package @babylonjs/react-native-windows
-        run: |
-          npm version --no-git-tag-version ${GITHUB_REF/refs\/tags\//}
-          npm publish --access public
-        working-directory: ./Package/Assembled-Windows
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+#      - name: Setup Node.js
```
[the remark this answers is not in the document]
CedricGuillemet: Yes, I have no secrets set on my repo so it would fail when trying to publish. I'm just cautious.

.github/workflows/typescript.yml
CedricGuillemet: Ho yes, I missed it!

Modules/@babylonjs/react-native-iosandroid/.gitignore
CedricGuillemet: No, it's added and same from package-lock.json: a new `react-native-iosandroid` folder is added with package-lock and gitignore. That's why there is a 22K lines more.

Modules/@babylonjs/react-native-iosandroid/.npmignore
```diff
@@ -0,0 +1,60 @@
+# OSX
```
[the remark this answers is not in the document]
CedricGuillemet: Same as package-lock and gitignore.

Modules/@babylonjs/react-native-iosandroid/package.json
CedricGuillemet: Indeed

Package/gulpfile.js
CedricGuillemet: Fixed.

CedricGuillemet:
> One other change I would include is to update the `VersionValidation.ts` file to check the `Platform.OS` and verify that you have the right binary package and version installed, and if not, emit an error. This will make it easy for people updating to the new @babylonjs/react-native package understand there are additional dependencies. It would probably also be good to have an exported variable from `VersionValidation.ts` to indicate whether the binary packages are present, and if not, have useEngine and EngineView no-op or something (and not explode).

The react-native-iosandroid and react-native-windows have a dependency on react-native with the exact same version. isn't it enough or do you want to check smth else?

Package/gulpfile.js
CedricGuillemet: I changed the function a bit so you can specify a react-native version and/or the package release version. Both are optional now.

Modules/@babylonjs/react-native/package.json
CedricGuillemet: Makes sense. Fixed.

SergioRZMasson:
Just tested the changes consuming the generated packages as well as building from the repo:

ios package for react native 0.64 - working
android package for react native 0.64 - working
ios package for react native 0.65 - working
android package for react native 0.65 - working

android app running from repo - working
ios app running from repo - working

windows app running from repo - working
windows package for react native 0.64 - builds with no error but shows message ("Could not initialize Babylon Native")
windows package for react native 0.65 - builds with no error but shows message ("Could not initialize Babylon Native")

Some users have reported on the forum that they were having the same issue when consuming windows package, I believe this was not caused by the changes in the package system. I will open an issue for us to track this windows package problem.

Modules/@babylonjs/react-native-iosandroid/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "@babylonjs/core": "^5.0.0-rc.4",
+    "@babylonjs/react-native": "version",
+    "react": ">=16.13.1",
+    "react-native": ">=0.63.1",
+    "react-native-permissions": "^3.0.0"
+  },
+  "devDependencies": {
+    "@babel/core": "^7.8.4",
+    "@babel/runtime": "^7.8.4",
+    "@babylonjs/core": "5.0.0-rc.4",
```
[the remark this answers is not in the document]
CedricGuillemet: Removed

Modules/@babylonjs/react-native/shared/XrContextHelper.h
```diff
@@ -1,179 +1,179 @@
...
-            xrContext = reinterpret_cast<IXrContextARKit*>(nativePtr);
-            return true;
-        }
-        
-        return false;
-    }
-}
-#endif
-#endif
+#pragma once
```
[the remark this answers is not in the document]
CedricGuillemet: No, the 2 Xr files were renamed 2 times (moved to react-native-iosandroid and then back to react-native). I reverted the 2 renames and the modification is gone.

Modules/@babylonjs/react-native/shared/XrAnchorHelper.h
```diff
@@ -1,229 +1,229 @@
...
-
-    bool TryDeclareNativeAnchor(Napi::Env env, const Napi::Value& session, ARAnchor* nativeAnchor, Napi::Value& xrAnchor)
-    {
-        uintptr_t nativeAnchorPtr{reinterpret_cast<uintptr_t>(nativeAnchor)};
-        return TryDeclareNativeAnchor(env, session, nativeAnchorPtr, xrAnchor);
-    }
-}
-#endif
-#endif
+#pragma once
```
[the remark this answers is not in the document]
CedricGuillemet: Reverted.

Modules/@babylonjs/react-native-windows/windows/CMakeLists.txt
CedricGuillemet: Before the change, both included CMakeLists.txt and submodules were in the same `react-native` folder. But now react-native-iosandroid contains the submodules folder and the cmakelists.txt is in `react-native`. 2 folders instead of just 1 before.

ryantrem: I see, I think the variable names are confusing now. Having `BABYLON_REACT_NATIVE_DIR` and `BABYLON_REACT_NATIVE_ANDROIDIOS` seems like it would be easier to understand.

CedricGuillemet: Fixed.

Modules/@babylonjs/react-native-iosandroid/package.json
CedricGuillemet: If it messes up the local build, then PR workflow would trigger an error when building the PG. Or am I missing smth?

Modules/@babylonjs/react-native-iosandroid/package.json
CedricGuillemet: I don't have a clear picture of 'good' dependency chain here. Feel free to update the PR or do another one if it's not good.

PR #366, opened 2022-03-28
new publishing GA

CedricGuillemet:
Github action that allows to publish NPM packages from any branch with any version


bghgary (approved):
Should consolidate after.

PR #384, opened 2022-03-31
Update Babylon Native to latest and Babylon.js to 5.0.0

bghgary:
Tested on Android from Windows


syntheticmagus (approved):

PolygonalSun:
Do we need to update the package-lock.json files in the PackageTest folders?

bghgary:
> Do we need to update the package-lock.json files in the PackageTest folders?

Not really since they need to be fixed up to use the new package scheme. @CedricGuillemet will hopefully take care of it soon.

PR #393, opened 2022-04-13
Package test 0.65.0

CedricGuillemet:
- ios/android/windows generated by react-native scripts
- metro config changed to support symlink
- packagetest-shared package containing app.tsx
- updated package.json w/ bjs deps

Tests:
- [x] Android
- [x] iOS
- [x] Windows

More PRs to come regarding 0.63.1, 0.64.0 and shared folder once this PR is merged.


Apps/PackageTest/0.65.0/index.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * @format
+ */
+
+import {AppRegistry} from 'react-native';
+import App from '../packagetest-shared/App';
```
ryantrem, line 6:
```suggestion
import { App } from 'packagetest-shared';
```

ryantrem (approved):
Maybe for a later change, but I think we should get rid of the PackageTest App.tsx, move playground-shared to Apps/playground-shared, add a package.json like you did for packagetest-shared, and use that playground-shared package in all of the Playground apps and all of the PackageTest apps. That way we can test all the various features through one common experience through all the test apps.

CedricGuillemet:
@ryantrem I'm wondering if we should publish the 'test' package containing app.tsx so we won't have to use symlinks and metroconfig would be more readable.

Apps/PackageTest/0.65.0/index.js
ryantrem: Oh, I see, because you are doing `export default App` rather than `export App`.

ryantrem:
> @ryantrem I'm wondering if we should publish the 'test' package containing app.tsx so we won't have to use symlinks and metroconfig would be more readable.

Publish the 'test' package containing app.tsx where? The reason it is setup this way is so that you can make changes to the common code (e.g. App.tsx in this case) and then just run the test apps, without having to worry about building or publishing packages. It makes a much tighter dev inner loop and so the workflow is much more efficient this way.

PR #409, opened 2022-05-17
Bugfix/android morph target

SergioRZMasson:
## Overiview

Sync with latest version of BabylonNative to get the changes that fix morph targets on Android. Also updated babylon.js packages used in the project to version 5.6.1 to get the fix from the js side and also to match PROTOCOL_VERSION (which has been increased to 6).

@babylonjs/react-native generated from this PR forward will only work with @babylonjs/core@5.6.1 or higher due to the change in PROTOCOL_VERSION

- Fixes #1049

### DIFF
--- a/Apps/PackageTest/0.63.1/package.json
+++ b/Apps/PackageTest/0.63.1/package.json
@@ -10,7 +10,7 @@
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
   },
   "dependencies": {
-    "@babylonjs/core": "5.2.0",
+    "@babylonjs/core": "5.6.1",
     "@babylonjs/react-native": "^0.4.0-alpha.47",
     "react": "16.13.1",
     "react-native": "0.63.1",
--- a/Apps/PackageTest/0.64.0/package-lock.json
+++ b/Apps/PackageTest/0.64.0/package-lock.json
@@ -8,7 +8,7 @@
       "name": "PackageTest",
       "version": "0.0.1",
       "dependencies": {
-        "@babylonjs/core": "5.2.0",
+        "@babylonjs/core": "5.6.1",
         "@babylonjs/react-native": "^0.4.0-alpha.47",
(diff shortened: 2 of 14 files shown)


Apps/PackageTest/0.65.0/package.json
```diff
@@ -11,7 +11,7 @@
     "windows": "react-native run-windows"
   },
   "dependencies": {
-    "@babylonjs/core": "^5.2.0",
```
ryantrem, line 14: Looks like the package-lock.json changes associated with this change are missing.

Apps/PackageTest/0.63.1/package.json
```diff
@@ -10,7 +10,7 @@
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
   },
   "dependencies": {
-    "@babylonjs/core": "5.2.0",
```
ryantrem, line 13: Looks like the package-lock.json changes associated with this change are missing.

Modules/@babylonjs/react-native-windows/package.json
```diff
@@ -24,7 +24,7 @@
   "licenseFilename": "LICENSE",
   "readmeFilename": "README.md",
   "peerDependencies": {
-    "@babylonjs/core": "^5.2.0",
```
ryantrem, line 27: Looks like the package-lock.json changes associated with this change are missing.

Apps/PackageTest/0.64.0/package-lock.json
```diff
@@ -8,7 +8,7 @@
       "name": "PackageTest",
       "version": "0.0.1",
       "dependencies": {
-        "@babylonjs/core": "5.2.0",
+        "@babylonjs/core": "5.6.1",
```
ryantrem, line 11: It seems weird that only one line in this file changed, and there isn't a hash update somewhere. Probably won't result in any issues, but kinda unexpected.

ryantrem (approved):

Issue #388, opened 2022-04-05
RN 0.68 Support


nmongiya:
When I am trying version 1.0.0 with React-Native 0.68
I am getting the below error
` WARN  Possible Unhandled Promise Rejection (id: 0):
TypeError: null is not an object (evaluating 'BabylonModule.initialize')
ensureInitialized$@http://localhost:8081/src/index.tsx.bundle?platform=android&dev=true&minify=false&app=XXX.XX.XX&modulesOnly=false&runModule=true:642083:60`

Here is my package.json
`"@babylonjs/core": "5.0.0",
  "@babylonjs/loaders": "5.0.0-alpha.60",
  "@babylonjs/materials": "5.0.0-alpha.60",
  "@babylonjs/react-native": "^1.0.1",`

**Expected behavior**
React-Native announced version 0.68 with same breaking changes 
https://reactnative.dev/blog/2022/03/30/version-068#breaking-changes-and-version-bumps
and upgrade Gradle to 7.3 is one of them> could that be the isssue?

**Other**
 - Platform: [Android]

ryantrem:
1. Make sure your first three Babylon.js packages are all the same version (5.0.0).
2. You need to install https://www.npmjs.com/package/@babylonjs/react-native-iosandroid-0-65 as well. Did you get a react error telling you to install this package?
3. We haven’t tested Babylon React Native 0.68 yet, so I’m not sure if it will work even with the above changes, but if not you should at least get a different error.

nmongiya:
1. Updated to make sure all three Babylon.js packages are all the same version (5.0.0).

`"@babylonjs/core": "5.0.0",
    "@babylonjs/loaders": "5.0.0",
    "@babylonjs/materials": "5.0.0",
    "@babylonjs/react-native": "^1.0.1",`
But still the same error.

2. It didn't report any error to install @babylonjs/react-native-iosandroid-0-65 but when I install it, it started to give me different errors.
 `1: Task failed with an exception.
-----------
* Where:
Build file 'XXXX\node_modules\@babylonjs\react-native-iosandroid-0-65\android\build.gradle' line: 27

* What went wrong:
A problem occurred evaluating project ':babylonjs_react-native-iosandroid-0-65'.
> Plugin with id 'maven' not found.`

I believe that is because node_modules\@babylonjs\react-native-iosandroid-0-65\android\build.gradle is pointing to maven
`maven {
[...]

PR #417, opened 2022-06-07
Remove Maven in Gradle script

### LINKED ISSUE #410: Remove unnecessary gradle configuration that impacts gradle compatibility
**Is your feature request related to a problem? Please describe.**
The Anrdoid build.gradle file seems to have some unnecessary configuration that reduces gradle compatibility.

**Describe the solution you'd like**
No build errors about unused parts of the gradle configuration in Babylon React Native from an app consuming it and using a different/newer version of gradle than the Babylon React Native Playground app.

**Describe alternatives you've considered**
We should go through the gradle file, especially this section: https://github.com/BabylonJS/BabylonReactNative/blob/master/Package/Android/build.gradle#L123-L166, and remove whatever parts are unnecessary. 

**Additional context**
Most recently, we discovered that the `installArchives` block seems to be unneeded yet breaks compatibility with gradle 7+.

CedricGuillemet:
fixes #410 

related https://github.com/BabylonJS/BabylonReactNative/issues/388

Tested with 0.65 on MacOS


ryantrem:
Don’t we need similar changes in https://github.com/BabylonJS/BabylonReactNative/blob/master/Package/Android/build.gradle?

pingumen96 (approved):
I tried this and everything works.

3GOMESz (approved):

CedricGuillemet:
> https://github.com/BabylonJS/BabylonReactNative/blob/master/Package/Android/build.gradle

Correct me if I'm wrong. The gradle you mentioned is used to build the package. If, for some reason, we had to build it with RN 0.68 (and 0.67), we would have to do the change. But 0.65 is compatible with 0.67 and 0.68, so it's not needed. I think it might be needed for 0.69 as it not ABI compatible.

ryantrem:
> > https://github.com/BabylonJS/BabylonReactNative/blob/master/Package/Android/build.gradle
> 
> Correct me if I'm wrong. The gradle you mentioned is used to build the package. If, for some reason, we had to build it with RN 0.68 (and 0.67), we would have to do the change. But 0.65 is compatible with 0.67 and 0.68, so it's not needed. I think it might be needed for 0.69 as it not ABI compatible.

The build.gradle file I referenced is the one that is included in the npm package for the Android build. See https://github.com/BabylonJS/BabylonReactNative/blob/master/Package/gulpfile.js#L278

CedricGuillemet:
> > > https://github.com/BabylonJS/BabylonReactNative/blob/master/Package/Android/build.gradle
> > 
> > 
> > Correct me if I'm wrong. The gradle you mentioned is used to build the package. If, for some reason, we had to build it with RN 0.68 (and 0.67), we would have to do the change. But 0.65 is compatible with 0.67 and 0.68, so it's not needed. I think it might be needed for 0.69 as it not ABI compatible.
> 
> The build.gradle file I referenced is the one that is included in the npm package for the Android build. See https://github.com/BabylonJS/BabylonReactNative/blob/master/Package/gulpfile.js#L278

My bad, you are right. Doing the change now.

Apps/Playground/0.64/ios/Playground.xcodeproj/project.pbxproj
```diff
@@ -403,7 +403,7 @@
 			);
 			inputPaths = (
 				"${PODS_ROOT}/Target Support Files/Pods-Playground/Pods-Playground-frameworks.sh",
-				"${PODS_XCFRAMEWORKS_BUILD_DIR}/OpenSSL/OpenSSL.framework/OpenSSL",
+				"${PODS_XCFRAMEWORKS_BUILD_DIR}/OpenSSL-Universal/OpenSSL.framework/OpenSSL",
```
ryantrem, line 406: Does this change relate to removing maven?

Apps/Playground/0.64/ios/Playground.xcodeproj/project.pbxproj
```diff
@@ -879,7 +879,7 @@
 				COPY_PHASE_STRIP = NO;
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
 				ENABLE_TESTABILITY = YES;
-				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = "arm64 ";
+				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = "";
```
ryantrem, line 882: Same question here.

Apps/Playground/0.64/ios/Podfile.lock
```diff
@@ -256,7 +256,7 @@ PODS:
   - React-jsinspector (0.64.1)
   - react-native-babylon (0.0.1):
     - React
-  - react-native-slider (4.0.0-rc.3):
+  - react-native-slider (4.2.3):
```
ryantrem, line 259: And here :)

CedricGuillemet:
@ryantrem Sergio did some changes because builds are failing for a few weeks now. @SergioRZMasson can you please elaborate on your changes and why it was failing actually ?

SergioRZMasson:
@ryantrem , we has some conflicts with the npm packages been used by 0.64. I had to update some of them to get npm install working again. The other changes might be due to the fact that I was using a arm64 mac, I think the changes made to project.pbxproj can be ignored, but the Podfile.lock is required since we have changed some dependencies versions.

ryantrem:
> @ryantrem , we has some conflicts with the npm packages been used by 0.64. I had to update some of them to get npm install working again. The other changes might be due to the fact that I was using a arm64 mac, I think the changes made to project.pbxproj can be ignored, but the Podfile.lock is required since we have changed some dependencies versions.

Ok, thanks for that context. Any specific reason those fixes are going into this PR, as opposed to being in a separate PR specifically about fixing the npm package ref issues?

ryantrem (approved):

3GOMESz (approved):

SergioRZMasson:
@ryantrem  It was mainly because it was failing the CI

PR #419, opened 2022-06-23
topMost property with perf improvement for Android

CedricGuillemet:
On Android, a partner was reporting slow updates/skipped frames when using transparency.
TextureView seems to be famous for being slow.
After some investigation, it's possible to do the same (textureView allows more things that are not usefull in this context) with SurfaceView/SurfaceHolder properties.

Change for iOS/Android. Another PR for Windows will come later.


Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
```diff
@@ -68,30 +68,26 @@ public void surfaceDestroyed(SurfaceHolder holder) {
...
-            if (this.opaqueSurfaceView == null) {
-                this.opaqueSurfaceView = new SurfaceView(this.getContext());
-                this.opaqueSurfaceView.setLayoutParams(EngineView.childViewLayoutParams);
-                this.opaqueSurfaceView.getHolder().addCallback(this);
-                this.addView(this.opaqueSurfaceView);
-            }
+            // ZOrder is not dynamic before Android 11. Recreate the surfaceView and set order before adding to the parent
+            // https://developer.android.com/reference/android/view/SurfaceView#setZOrderOnTop(boolean)
+            this.surfaceView.setZOrderOnTop(true);    // necessary
+            surfaceHolder.setFormat(PixelFormat.TRANSPARENT);
```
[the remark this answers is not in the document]
CedricGuillemet: We can add a 'transparencyMode' property that will only change behavior on Android.

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
```diff
@@ -26,17 +27,19 @@
 
 public final class EngineView extends FrameLayout implements SurfaceHolder.Callback, TextureView.SurfaceTextureListener, View.OnTouchListener {
     private static final FrameLayout.LayoutParams childViewLayoutParams = new FrameLayout.LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.MATCH_PARENT);
-    private TextureView transparentTextureView;
-    private Surface transparentSurface = null;
-    private SurfaceView opaqueSurfaceView = null;
+    private SurfaceView surfaceView = null;
```
ryantrem, line 30: To me the names were more clear the way they were before (`opaqueSurfaceView` and `transparentTextureView`). Unless you object will you change them back (and restore the original order to simplify the diff for this PR)?

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
```diff
@@ -67,34 +70,58 @@ public void surfaceDestroyed(SurfaceHolder holder) {
...
+        setIsTransparentAndTopMost(this.isTransparent, topMost);
+    }
     // ------------------------------------
     // TextureView related
-
     public void setIsTransparent(Boolean isTransparent) {
+        setIsTransparentAndTopMost(isTransparent, this.topMost);
+    }
+
+    public void setIsTransparentAndTopMost(Boolean isTransparent, Boolean topMost) {
```
ryantrem, line 83: This function is only used within this class, right? Not part of the contract I think?

```suggestion
    private void setIsTransparentAndTopMost(Boolean isTransparent, Boolean topMost) {
```

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
```diff
@@ -67,34 +70,58 @@ public void surfaceDestroyed(SurfaceHolder holder) {
...
+                this.surfaceView = new SurfaceView(this.getContext());
+                this.surfaceView.setLayoutParams(EngineView.childViewLayoutParams);
+                SurfaceHolder surfaceHolder = this.surfaceView.getHolder();
+
+                // ZOrder is not dynamic before Android 11. Recreate the surfaceView and set order before adding to the parent
+                // https://developer.android.com/reference/android/view/SurfaceView#setZOrderOnTop(boolean)
+                this.surfaceView.setZOrderOnTop(true);    // necessary
+                surfaceHolder.setFormat(PixelFormat.TRANSPARENT);
+                this.addView(this.surfaceView);
+            } else if (this.textureView == null) {
```
ryantrem, line 103: Won't the textureView always be null at this point since it was unconditionally cleared out on lines 88-91?

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
```diff
@@ -67,34 +70,58 @@ public void surfaceDestroyed(SurfaceHolder holder) {
...
-            if (this.transparentTextureView != null) {
-                this.transparentTextureView.setVisibility(View.GONE);
-                this.transparentTextureView = null;
-            }
-            if (this.opaqueSurfaceView == null) {
-                this.opaqueSurfaceView = new SurfaceView(this.getContext());
-                this.opaqueSurfaceView.setLayoutParams(EngineView.childViewLayoutParams);
-                this.opaqueSurfaceView.getHolder().addCallback(this);
-                this.addView(this.opaqueSurfaceView);
+            if (this.surfaceView == null) {
```
ryantrem, line 111: Similar question here.

Modules/@babylonjs/react-native/EngineView.tsx
```diff
@@ -9,6 +9,7 @@ export interface EngineViewProps extends ViewProps {
     camera?: Camera;
     displayFrameRate?: boolean;
     isTransparent?: boolean;
+    topMost?: boolean;
```
ryantrem, line 12: Should this be `isTopMost` to match the naming scheme for transparency (`isTransparent`)?

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
```diff
@@ -67,34 +70,58 @@ public void surfaceDestroyed(SurfaceHolder holder) {
...
+    public void setIsTransparentAndTopMost(Boolean isTransparent, Boolean topMost) {
+        if (this.surfaceView != null) {
+            this.surfaceView.setVisibility(View.GONE);
+            this.surfaceView = null;
+        }
+        if (this.textureView != null) {
+            this.textureView.setVisibility(View.GONE);
+            this.textureView = null;
+        }
         if (isTransparent) {
```
ryantrem, line 92: These conditions seems like they could be simplified. Conceptually can't it just be:
```ts
if (isTransparent && !topMost) {
  // use TextureView
} else {
  // use SurfaceView (configured according to isTransparent and topMost)
}
```

It seems like this would simplify the logic and remove some code duplication, which should also reduce opportunities for coding errors. Speaking of which, isn't the first case of creating the `SurfaceView` missing the `addCallback` which deals with resizing?

Modules/@babylonjs/react-native/EngineView.tsx
CedricGuillemet: Makes sense.

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
CedricGuillemet: I can make it more clear for the textureView to rename it transparentTextureView. But the surfaceView can be opaque or transparent.

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
CedricGuillemet: Indeed! It's better.

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
CedricGuillemet: Fixed.

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
CedricGuillemet: Fixed as well.

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
ryantrem: Ok, maybe if they just stay in the same order then it will be more clear.

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
```diff
@@ -67,34 +71,50 @@ public void surfaceDestroyed(SurfaceHolder holder) {
...
-                this.transparentTextureView = new TextureView(this.getContext());
-                this.transparentTextureView.setLayoutParams(EngineView.childViewLayoutParams);
-                this.transparentTextureView.setSurfaceTextureListener(this);
-                this.transparentTextureView.setOpaque(false);
-                this.addView(this.transparentTextureView);
-            }
+        setIsTransparentAndIsTopMost(isTransparent, this.isTopMost);
+    }
+
+    private void setIsTransparentAndIsTopMost(Boolean isTransparent, Boolean isTopMost) {
```
ryantrem, line 84: Should we be checking the inputs against the current values and not recreating views when unnecessary?

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
```diff
@@ -67,34 +71,50 @@ public void surfaceDestroyed(SurfaceHolder holder) {
...
+            SurfaceHolder surfaceHolder = this.surfaceView.getHolder();
+            if (isTransparent) {
+                surfaceHolder.setFormat(PixelFormat.TRANSPARENT);
             }
-            if (this.opaqueSurfaceView == null) {
-                this.opaqueSurfaceView = new SurfaceView(this.getContext());
-                this.opaqueSurfaceView.setLayoutParams(EngineView.childViewLayoutParams);
-                this.opaqueSurfaceView.getHolder().addCallback(this);
-                this.addView(this.opaqueSurfaceView);
+            if (isTopMost || isTransparent) {
```
ryantrem, line 106: This can just be `if (isTopMost)` right? Otherwise we are saying we want to catch the case where `isTopMost` is false and `isTransparent` is true, but that's the first condition of the higher level logic that would use `TextureView`.

CedricGuillemet: Fixed.

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
CedricGuillemet: Fixed.

PR #421, opened 2022-06-28
view MSAA

CedricGuillemet:
Expose MSAA settings as a view property

Tests:

- [x] Android 0.65
- [x] Windows
- [x] iOS


Package/iOS/CMakeLists.txt
```diff
@@ -5,6 +5,14 @@ include(${CMAKE_CURRENT_SOURCE_DIR}/../../Apps/Playground/Playground/node_module
...
     astc-codec
     astc
+	etc1
+	etc2
+	nvtt
+	squish
+	pvrtc
+	iqa
+	edtaa3
+	tinyexr
```
ryantrem, line 15:
```suggestion
    etc1
    etc2
    nvtt
    squish
    pvrtc
    iqa
    edtaa3
    tinyexr
```

ryantrem, line 15: I think these need to be added to `Modules/@babylonjs/react-native-iosandroid/react-native-babylon.podspec`, though I would have expected this to fail to link on iOS without that change. Is that not happening?

ryantrem, line 15: These are additional libs we are now linking, right? If so they need to be added to gulpfile.js as well. I would definitely make sure to test the packages any time we add new dependencies.

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
```diff
@@ -261,6 +272,18 @@ namespace BabylonNative
...
 
+    void UpdateMSAA(uint8_t value)
+    {
+        if (auto nativeModule{ g_nativeModule.lock() })
+        {
+            nativeModule->UpdateMSAA(value);
+        }
+        else
+        {
+            throw std::runtime_error{ "UpdateView must not be called before Initialize." };
```
ryantrem, line 283:
```suggestion
            throw std::runtime_error{ "UpdateMSAA must not be called before Initialize." };
```

Modules/@babylonjs/react-native/EngineView.tsx
```diff
@@ -26,7 +27,8 @@ export const EngineView: FunctionComponent<EngineViewProps> = (props: EngineView
     const [sceneStats, setSceneStats] = useState<SceneStats>();
     const engineViewRef = useRef<Component<NativeEngineViewProps>>(null);
     const snapshotPromise = useRef<{ promise: Promise<string>, resolve: (data: string) => void }>();
-    const isTransparent = props.isTransparent || false
+    const isTransparent = props.isTransparent || false;
+    const MSAA = props.MSAA || 0;
```
ryantrem, line 31:
```suggestion
    const isTransparent = props.isTransparent ?? false;
    const MSAA = props.MSAA ?? 0;
```

Modules/@babylonjs/react-native/EngineView.tsx
```diff
@@ -9,6 +9,7 @@ export interface EngineViewProps extends ViewProps {
     camera?: Camera;
     displayFrameRate?: boolean;
     isTransparent?: boolean;
+    MSAA: number;
```
ryantrem, line 12: Maybe name this `antiAliasing` to avoid acronyms in the public API?

ryantrem, line 12: Narrow this from any number to specific valid numbers, and also make it optional:
```suggestion
    antiAliasing?: 0 | 1 | 2 | 4 | 8 | 16;
```

Package/iOS/CMakeLists.txt
CedricGuillemet: No, it happens only when trying to link the app. I believed PG was built on the CI.

CedricGuillemet: Done.

ryantrem (approved):

PR #429, opened 2022-07-20
Alpha premultiplied backbuffer Windows

CedricGuillemet:
follow up to https://github.com/BabylonJS/BabylonNative/pull/1100
Change alpha premultiplied back buffer depending on transparency settings. 
`isTopMost` property not implemented as it's not mandatory for our partner and it seems like it needs deeper understanding and (a lot) more work.
PR will be updated with submodule update when BabylonNative PR is merged.

### DIFF
--- a/Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
+++ b/Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
@@ -183,7 +183,7 @@ namespace winrt::BabylonReactNative::implementation {
             if (propertyName == "isTransparent")
             {
                 bool isTransparent = propertyValue.AsBoolean();
-                this->CompositeMode(isTransparent ? ElementCompositeMode::MinBlend : ElementCompositeMode::Inherit);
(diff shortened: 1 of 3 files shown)


Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
```diff
@@ -183,7 +183,7 @@ namespace winrt::BabylonReactNative::implementation {
             if (propertyName == "isTransparent")
             {
                 bool isTransparent = propertyValue.AsBoolean();
-                this->CompositeMode(isTransparent ? ElementCompositeMode::MinBlend : ElementCompositeMode::Inherit);
+                BabylonNative::UpdateAlphaPremultiplied(isTransparent);
```
ryantrem, line 186: So on Windows, when the isTransparent flag is true, we need to set the platform agnostic alpha premultiplied to true as well. Is this only for Windows? Will the logic be different for the other platforms (iOS and Android)?

ryantrem (approved):

PR #432, opened 2022-07-22
Fix BabylonModule TypeError

### LINKED ISSUE #431: TypeError: null is not an object (evaluating 'BabylonModule.initialize')
**Describe the bug**
A big warn every time applications starts...

Probably because NativeModule can be null and TypeScript emphasizes this

https://github.com/BabylonJS/BabylonReactNative/blob/1ed3e5becab8a78e8229973be057393560d47ed9/Modules/%40babylonjs/react-native/BabylonModule.ts#L22

**To Reproduce**
N/A

**Other**
 - Platform: Windows 10, Android

runtothedoor:
fix #431

### DIFF
--- a/Modules/@babylonjs/react-native/BabylonModule.ts
+++ b/Modules/@babylonjs/react-native/BabylonModule.ts
@@ -12,7 +12,7 @@ const BabylonModule: {
     resetView(): Promise<void>;
 } = NativeModules.BabylonModule;
 
-export async function ensureInitialized() {
+export async function ensureInitialized(): Promise<boolean | undefined> {
     if (isRemoteDebuggingEnabled) {
         // When remote debugging is enabled, JavaScript runs on the debugging host machine, not on the device where the app is running.
         // JSI (which Babylon Native uses heavily) can not work in this mode. In the future, this debugging mode will be phased out as it is incompatible with TurboModules for the same reason.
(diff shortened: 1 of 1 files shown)


Modules/@babylonjs/react-native/BabylonModule.ts
```diff
@@ -18,9 +18,13 @@ export async function ensureInitialized(): Promise<boolean> {
...
     } else {
-        // This does the first stage of Babylon Native initialization, including creating the BabylonNative JSI object.
-        await BabylonModule.initialize();
-        return true;
+        if(BabylonModule){ 
+            // This does the first stage of Babylon Native initialization, including creating the BabylonNative JSI object.
+            await BabylonModule.initialize();
+            return true;
+        } else{
+            return false;
```
runtothedoor: My mistake here, it shouldn't return false in this case

runtothedoor:
Sorry for the lots of commits that result in a few lines change, but depending on the code different errors appeared and now I believe everything is ok...

Modules/@babylonjs/react-native/BabylonModule.ts
```diff
@@ -22,9 +22,7 @@ export async function ensureInitialized(): Promise<boolean> {
             // This does the first stage of Babylon Native initialization, including creating the BabylonNative JSI object.
             await BabylonModule.initialize();
             return true;
-        } catch {
-            throw Error('[ERROR] Could not find "BabylonModule" in NativeModules!');
```
runtothedoor: Line removed as it appears in console on first load or refresh

PR #433, opened 2022-07-26
Enable native camera plugin

CoPrez:
This change enables the BN NativeCamera plugin for the BRN project. On Android the NativeCamera plugin uses a compile time check ([here](https://github.com/BabylonJS/BabylonNative/blob/7c6d12d0e13052ae1cc95c0fc0ca54d117432559/Plugins/NativeCamera/Source/Android/NativeCameraImpl.cpp#L156)) to only support projects targeting a min SDK of 24 or higher. Since BRN includes precompiled binaries for BN that means this change also bumps the BRN playground apps up to min SDK 24 so that the NativeCamera plugin compiles with the proper implementation.

I used the PackageTest projects (which still have a min SDK of 21) to ensure including the pre-compiled binaries against min SDK 24 didn't break anything in the consuming apps.

A better long term fix would be to update the NativeCamera's implementation to use a runtime check of the API level rather than a compile time check.


ryantrem (approved):

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/BabylonReactNative.vcxproj
```diff
@@ -252,6 +254,7 @@
         JsRuntime.lib;
         MachineIndependent.lib;
         napi.lib;
+        NativeCamera.lib
```
ryantrem, line 257:
```suggestion
        NativeCamera.lib;

```

CoPrez: Thanks, added in the missing `;`

ryantrem (approved):

Issue #428, opened 2022-07-19
React Native 0.69 support


ryantrem:
**Is your feature request related to a problem? Please describe.**
React Native 0.69 was recently released, but has a [JSI ABI breaking change](https://github.com/facebook/hermes/commit/2b6d408980d7f23f50602fd88169c8a9881592a6). This means the existing BRN packages don't work with RN 0.69.

**Describe the solution you'd like**
Add another RN flavor for 0.69 and publish the corresponding package.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

saidmoya12:
Please add support to RN 68 too

ryantrem:
@CedricGuillemet are you planning to publish a package with the 0.68 fix (the maven issue )?

CedricGuillemet:
I've published the 1.3.0-preview (being tested with 1 of our partners) : https://www.npmjs.com/package/@babylonjs/react-native-iosandroid-0-65/v/1.3.0-preview

it contains the gradle/maven fix.

saidmoya12:
Maven works but useEngine return null always, some idea?

Thanks

ryantrem:
@saidmoya12 will you start a thread on the [forum](https://forum.babylonjs.com) and include some or all of your code so we can see what you are trying to do?

runtothedoor:
Any updates?

theonetheycallneo:
I'm reposting this here (from 0.68 issue) for 0.69 as well:

Does anyone have a workaround for RN 0.68 with BabylonJs for iOS?

```
2022-08-02 11:55:30.626774-0400 Wallet[68917:2936056] [javascript] Possible Unhandled Promise Rejection (id: 0):
TypeError: Cannot read property 'initialize' of null
TypeError: Cannot read property 'initialize' of null
    at ensureInitialized$ (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.vatomcorp.viewer:653325:71)
    at call (native)
    at tryCatch (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.vatomcorp.viewer:24153:23)
    at anonymous (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=com.vatomcorp.viewer:24133:34)
[...]

PR #442, opened 2022-08-06
Create new package to support React Native 0.69

Alex-MSFT:
**Describe the change**

React Native 0.69 contains some ABI breaking changes, which require creating a new Babylon React Native package to support these scenarios.  This PR introduces a new Playground and build flavor to unblock apps from updating to 0.69.*.

See: #428

**Testing**

Built playground locally on Windows, android, and iOS to verify they work.  Tested new PackageTest project for Android and iOS with local builds.

Will add Windows support to PackageTest 0.69.0 once the first package is published.


Apps/PackageTest/0.69.0/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "android": "react-native run-android",
+    "ios": "react-native run-ios",
+    "start": "react-native start",
+    "test": "jest",
+    "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
+  },
+  "dependencies": {
+    "@babylonjs/core": "^5.6.1",
+    "@babylonjs/react-native": "^1.0.1",
+    "@babylonjs/react-native-iosandroid-0-69": "^1.0.1",
```
Alex-MSFT, line 15: Windows support will be added to PackageTest after the first 0-69 package is released.

.github/workflows/ios_android.yml
```diff
@@ -22,6 +26,11 @@ jobs:
         uses: jwlawson/actions-setup-cmake@v1.8
         with:
           cmake-version: '3.19.6' # See https://gitlab.kitware.com/cmake/cmake/-/issues/22021
+      - name: Set up JDK
+        uses: actions/setup-java@v2
+        with:
+          distribution: 'zulu'
+          java-version: ${{ inputs.java-version }}
```
ryantrem, line 33: Does the build fail without this or something? I'm trying to understand the implication... does this mean ADO build agents don't work with newer Android projects or something?

Apps/PackageTest/0.69.0/metro.config.js
```diff
@@ -0,0 +1,108 @@
+/**
```
ryantrem, line 1: We don't need all the customizations for dealing with sym links since PackageTest is specifically for testing the built package. Looks like this got messed up in the 0.65 version of PackageTest. I'd say we should at least fix it for this file (I believe the default metro.config.js should be fine), and ideally fix it for 0.65 too, though we could do that later if needed.

Apps/Playground/playground-shared/App.tsx
```diff
@@ -140,7 +140,7 @@ const EngineScreen: FunctionComponent<ViewProps> = (props: ViewProps) => {
                 <Image style={{flex: 1}} source={{uri: snapshotData }} />
               </View>
             }
-            <EngineView camera={camera} onInitialized={onInitialized} displayFrameRate={true} antiAliasing={2}/>
```
ryantrem, line 143: Can you try with this enabled again? I think @CedricGuillemet made a fix for this in BGFX a few days ago, and I think that has made it's way to master in BRN as well.

ryantrem:
Did you take notes on all the steps involved in adding support for a new RN version? Would you be able to update the root readme.md with instructions on how to do this to make it easier in the future?

.github/workflows/ios_android.yml
Alex-MSFT: Yes by default the build machines were pointing to JDK 8, so omitting this step was causing a build failer here is the documentation about this task https://github.com/actions/setup-java

Apps/Playground/playground-shared/App.tsx
Alex-MSFT: Verified this works now

ryantrem (approved):

runtothedoor:
@Alex-MSFT

Issue #440, opened 2022-08-03
VideoTexture.CreateFromWebCam doesn't properly request camera permissions


CoPrez:
Due to the complexity of requesting permissions on different platforms (particularly Android) the current implementation of `VideoTexture.CreateFromWebCam` fails on Android if it has to prompt the user for permissions. The prompt shows up, but the VideoTexture doesn't load because it's not properly awaiting the camera permission. It does however work if the camera permission was previously granted to the app.

A work-around was added specifically for the camera permission when entering XR:
https://github.com/BabylonJS/BabylonReactNative/blob/410290dced6ba6d0b0a8df13145acc76a94acfc6/Modules/%40babylonjs/react-native/EngineHook.ts#L29

A similar work-around should be added for `VideoTexture.CreateFromWebCam` as well.

PR #450, opened 2022-08-17
VideoTexture WebCam Permission Check

CoPrez:
**Describe the change**
This PR addresses #440 - VideoTexture.CreateFromWebCam doesn't properly request camera permissions

The VideoTexture.CreateFromWebCamAsync method has an internal camera permission check, but due to the complexity of requesting permissions in a React-Native app it does not work in the RN project. A similar work around was created for the XR camera permission check.
**Testing**

I tested this change on Android by modifying the Playground apps camera permission in the OS setting and ensuring it would properly request the permission at runtime.

### DIFF
--- a/Modules/@babylonjs/react-native/EngineHook.ts
+++ b/Modules/@babylonjs/react-native/EngineHook.ts
@@ -56,8 +56,8 @@ async function requestCameraPermissionAsync() : Promise<void> {
 // 1. React Native does not provide a way to hook into the permissions request result (at least on Android).
 // 2. If it is done on the native side, then we need one implementation per platform.
 {
-    const originalInitializeSessionAsync: (...args: any[]) => Promise<any> = WebXRSessionManager.prototype.initializeSessionAsync;
-    WebXRSessionManager.prototype.initializeSessionAsync = async function (...args: any[]): Promise<any> {
+    const originalInitializeSessionAsync = WebXRSessionManager.prototype.initializeSessionAsync;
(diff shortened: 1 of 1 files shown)


Alex-MSFT (approved):

Modules/@babylonjs/react-native/EngineHook.ts
```diff
@@ -34,33 +65,21 @@ class DOMException {
...
-        if (permissionStatus == "denied")
-        {
-            permissionStatus = await request(cameraPermission);
-        }
+// Override the VideoTexture.CreateFromWebCamAsync to insert a camera permissions request. It would be cleaner to do this directly in the NativeCamera implementation, but there are a couple problems with that:
+// 1. React Native does not provide a way to hook into the permissions request result (at least on Android).
+// 2. If it is done on the native side, then we need one implementation per platform.
+{
+    const originalCreateFromWebCamAsync: (...args: any[]) => Promise<any> = VideoTexture.CreateFromWebCamAsync;
+    VideoTexture.CreateFromWebCamAsync = async function (...args: any[]): Promise<any> {
```
ryantrem, line 79: I know this is copying the existing pattern, but I think we could make these much more strongly typed and at the same time simplify it:
```suggestion
    const originalCreateFromWebCamAsync = VideoTexture.CreateFromWebCamAsync;
    VideoTexture.CreateFromWebCamAsync = async function (...args: Parameters<typeof originalCreateFromWebCamAsync>): ReturnType<typeof originalCreateFromWebCamAsync> {
```

ryantrem (approved):

Modules/@babylonjs/react-native/EngineHook.ts
CoPrez: I like it. I updated both method overrides to use the same type safe declarations.

PR #454, opened 2022-09-06
Fix fast refresh for PG

CedricGuillemet:
Before this PR: initPromise was created in resetView. But the promise was checked here https://github.com/BabylonJS/BabylonReactNative/blob/fb054db8f457a369d4f073ecc062e5d2158a92bc/Modules/%40babylonjs/react-native/ReactNativeEngine.ts#L20 before it was created again. As it was resolved, bgfx was assumed to be available. JS continues its work until bgfx is really shutdown and we get a crash (partner experience)

fix: When asked to reset the view from the JS side, reset the Promise. 

Note for fast refresh:
When a fast refresh is asked, the view is unchanged meaning `UpdateView` is not called. As the promise is resolved in `UpdateView`, rendering will stop.
To fix that, I've added a `m_resetDone` boolean that gets true when `resetView` is called and rendering is disabled. As rendering continues, this boolean is check to effectively `UpdateView` with the last know window configuration.

### DIFF
--- a/Apps/Playground/playground-shared/App.tsx
+++ b/Apps/Playground/playground-shared/App.tsx
@@ -107,7 +107,7 @@ const EngineScreen: FunctionComponent<ViewProps> = (props: ViewProps) => {
             setTrackingState(newTrackingState);
           });
 
-          // TODO: Figure out why getFrontPosition stopped working
+          // TODO: Figure out why getFrontPosition stopped working.
           //box.position = (scene.activeCamera as TargetCamera).getFrontPosition(2);
           const cameraRay = scene.activeCamera!.getForwardRay(1);
           rootNode.position = cameraRay.origin.add(cameraRay.direction.scale(cameraRay.length));
--- a/Apps/Playground/playground-shared/App.tsx
+++ b/Apps/Playground/playground-shared/App.tsx
@@ -127,7 +127,6 @@ const EngineScreen: FunctionComponent<ViewProps> = (props: ViewProps) => {
     }
   }, [engineViewCallbacks]);
 
-  // @refresh reset
   return (
     <>
       <View style={props.style}>
--- a/Apps/Playground/playground-shared/App.tsx
+++ b/Apps/Playground/playground-shared/App.tsx
@@ -107,7 +107,7 @@ const EngineScreen: FunctionComponent<ViewProps> = (props: ViewProps) => {
             setTrackingState(newTrackingState);
           });
 
-          // TODO: Figure out why getFrontPosition stopped working.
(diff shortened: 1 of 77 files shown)


Modules/@babylonjs/react-native/shared/BabylonNative.cpp
```diff
@@ -105,10 +104,15 @@ namespace BabylonNative
...
 
-            m_jsDispatcher([this]()
+            if (!m_isRenderingEnabled)
             {
-                m_resolveInitPromise();
-            });
+                m_jsDispatcher([this]()
+                    {
+                        m_resolveInitPromise();
+                        CreateInitPromise();
```
CedricGuillemet, line 112: A new promise is created after it's been resolved instead of when view is reset. Also, it's only done once thanks to `m_isRenderingEnabled` check.

Apps/Playground/playground-shared/App.tsx
```diff
@@ -127,6 +127,7 @@ const EngineScreen: FunctionComponent<ViewProps> = (props: ViewProps) => {
     }
   }, [engineViewCallbacks]);
 
+  // @refresh reset
```
ryantrem, line 130: Couldn't this just be done inside EngineView.tsx?

ryantrem:
Do we really want to tear down and re-create bgfx/graphics on fast refresh? Fast refresh is meant to enable a very fast dev inner loop, and be able to quickly show incremental changes. If we re-create bgfx/graphics, is that noticeably slower and/or disruptive to the dev inner loop?

Also, regarding the UpdateView not getting called after a fast refresh... the way the code is currently written, creating a ReactNativeEngine instance must be deferred until graphics is initialized, which in turn must be deferred until there is a window. But if in the case of fast refresh we already have a window (and it is reusable), then it seems like we just need to adjust the code to have the graphics initialization be triggered by something else and see there is already a window to use. I could see doing something like:
- Tracking an arcana task inside BabylonNative.cpp that is completed when a window is eventually provided (via UpdateView)
- Tracking an arcana task inside BabylonNative.cpp for initializing graphics, which could be triggered by the call to retrieve `initializationPromise`, and would await the window task above

Then if we don't have a window yet, async graphics init will need to wait. If we already have a window, then it will just continue on and init if needed.

Apps/Playground/playground-shared/App.tsx
CedricGuillemet: I tried but it doesn't work.

CedricGuillemet:
@ryantrem yes, fast refresh could be faster without bgfx re-init. It needs a separate issue/PR.

PR #460, opened 2022-09-13
Android ZOrder for transparent surfaceView

CedricGuillemet:
As discussed offline, add a new `androidView` property and remove `isTopMost`

### DIFF
--- a/Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
+++ b/Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
@@ -34,14 +34,13 @@ public final class EngineView extends FrameLayout implements SurfaceHolder.Callb
     private SurfaceView xrSurfaceView;
     private boolean isTransparent = false;
     private boolean isTopMost = false;
-    private boolean isOverlay = false;
     private final EventDispatcher reactEventDispatcher;
     private Runnable renderRunnable;
 
     public EngineView(ReactContext reactContext) {
         super(reactContext);
 
-        this.setViewFlags(false, false, false);
+        this.setIsTransparentAndIsTopMost(false, false);
 
         this.xrSurfaceView = new SurfaceView(reactContext);
         this.xrSurfaceView.setLayoutParams(childViewLayoutParams);
@@ -74,20 +73,16 @@ public final class EngineView extends FrameLayout implements SurfaceHolder.Callb
     }
 
     public void setIsTopMost(Boolean isTopMost) {
-        setViewFlags(this.isTransparent, isTopMost, this.isOverlay);
-    }
-
-    public void setIsOverlay(Boolean isOverlay) {
-        setViewFlags(this.isTransparent, this.isTopMost, isOverlay);
+        setIsTransparentAndIsTopMost(this.isTransparent, isTopMost);
     }
     // ------------------------------------
     // TextureView related
     public void setIsTransparent(Boolean isTransparent) {
-        setViewFlags(isTransparent, this.isTopMost, this.isOverlay);
+        setIsTransparentAndIsTopMost(isTransparent, this.isTopMost);
     }
 
-    private void setViewFlags(Boolean isTransparent, Boolean isTopMost, Boolean isOverlay) {
-        if (this.isTransparent == isTransparent && this.isTopMost == isTopMost && this.isOverlay == isOverlay &&
+    private void setIsTransparentAndIsTopMost(Boolean isTransparent, Boolean isTopMost) {
+        if (this.isTransparent == isTransparent && this.isTopMost == isTopMost &&
                 (this.surfaceView != null || this.transparentTextureView != null)) {
             return;
         }
@@ -99,7 +94,7 @@ public final class EngineView extends FrameLayout implements SurfaceHolder.Callb
             this.transparentTextureView.setVisibility(View.GONE);
             this.transparentTextureView = null;
         }
-        if (isTransparent && !isTopMost && !isOverlay) {
+        if (isTransparent && !isTopMost) {
             this.transparentTextureView = new TextureView(this.getContext());
             this.transparentTextureView.setLayoutParams(EngineView.childViewLayoutParams);
             this.transparentTextureView.setSurfaceTextureListener(this);
@@ -112,12 +107,10 @@ public final class EngineView extends FrameLayout implements SurfaceHolder.Callb
             if (isTransparent) {
                 surfaceHolder.setFormat(PixelFormat.TRANSPARENT);
             }
-            if (isOverlay) {
+            if (isTopMost) {
+                // setZOrderMediaOverlay  is not dynamic before Android 11. Recreate the surfaceView and set order before adding to the parent
+                // https://developer.android.com/reference/android/view/SurfaceView#setZOrderMediaOverlay(boolean)
                 this.surfaceView.setZOrderMediaOverlay(true);
-            } else if (isTopMost) {
-                // ZOrder is not dynamic before Android 11. Recreate the surfaceView and set order before adding to the parent
-                // https://developer.android.com/reference/android/view/SurfaceView#setZOrderOnTop(boolean)
-                this.surfaceView.setZOrderOnTop(true);
             }
             surfaceHolder.addCallback(this);
             this.addView(this.surfaceView);
@@ -125,7 +118,6 @@ public final class EngineView extends FrameLayout implements SurfaceHolder.Callb
 
         this.isTransparent = isTransparent;
         this.isTopMost = isTopMost;
-        this.isOverlay = isOverlay;
 
         // xr view needs to be on top of views that might be created after it.
         if (this.xrSurfaceView != null) {
--- a/Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
+++ b/Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
@@ -34,13 +34,14 @@ public final class EngineView extends FrameLayout implements SurfaceHolder.Callb
     private SurfaceView xrSurfaceView;
     private boolean isTransparent = false;
     private boolean isTopMost = false;
+    private boolean isOverlay = false;
     private final EventDispatcher reactEventDispatcher;
     private Runnable renderRunnable;
 
     public EngineView(ReactContext reactContext) {
         super(reactContext);
 
-        this.setIsTransparentAndIsTopMost(false, false);
+        this.setViewFlags(false, false, false);
 
         this.xrSurfaceView = new SurfaceView(reactContext);
         this.xrSurfaceView.setLayoutParams(childViewLayoutParams);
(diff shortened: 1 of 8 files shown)


bghgary:
I couldn't find any info on this. What is the difference between these two functions?

CedricGuillemet:
> I couldn't find any more info on this. What is the difference between these two functions in practice?

`setzorderontop` will make the surfaceView on top of everything that's under (surfaceView, textureView,... anything), independently of view hierarchy.
`setzordermediaoverlay` will composite the surfaceView on top of other surfaceViews

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
```diff
@@ -108,9 +108,9 @@ private void setIsTransparentAndIsTopMost(Boolean isTransparent, Boolean isTopMo
                 surfaceHolder.setFormat(PixelFormat.TRANSPARENT);
             }
             if (isTopMost) {
-                // ZOrder is not dynamic before Android 11. Recreate the surfaceView and set order before adding to the parent
-                // https://developer.android.com/reference/android/view/SurfaceView#setZOrderOnTop(boolean)
-                this.surfaceView.setZOrderOnTop(true);
+                // setZOrderMediaOverlay  is not dynamic before Android 11. Recreate the surfaceView and set order before adding to the parent
```
ryantrem, line 111:
```suggestion
                // setZOrderMediaOverlay is not dynamic before Android 11. Recreate the surfaceView and set order before adding to the parent
```

ryantrem:
`setZOrderMediaOverlay` says *"the surface view's surface is placed on top of another regular surface view in the window (but still behind the window itself)"*. Doesn't that mean it is not top most (i.e. it is still behind the window itself and therefore behind all the normal UI)?

CedricGuillemet:
> `setZOrderMediaOverlay` says _"the surface view's surface is placed on top of another regular surface view in the window (but still behind the window itself)"_. Doesn't that mean it is not top most (i.e. it is still behind the window itself and therefore behind all the normal UI)?

Yes, it's not topmost on Android anymore (as it's still on iOS). It's topMost on other surfaceViews. It's platform specific behaviors. I don't see any obvious name for that. Or even if just a name change is enough.

bghgary:
It seems like this flag is no longer consistent across platforms with this change. That seems bad. I wonder if it makes sense to expose the underlying platform view and the application can do whatever it wants with it?

CedricGuillemet:
> It seems like this flag is no longer consistent across platforms with this change. That seems bad. I wonder if it makes sense to expose the underlying platform view and the application can do whatever it wants with it?

Do you mean exposing a MTKView or a surfaceView to JS? Or asking user to add a bridge between the native view and JS?

bghgary:
> Do you mean exposing a MTKView or a surfaceView to JS? Or asking user to add a bridge between the native view and JS?

I'm not sure. Is there already an existing JS visible interface for talking to the native view? I'm guessing no.

ryantrem:
What I've seen in other RN components is having props that are platform specific and documented as such. In this specific case, what is the behavior you are trying to achieve? It seems like maybe there is a SurfaceView, and then the BRN EngineView is supposed to go on top of that, but then UI is supposed to go on top of everything? Although that's not the behavior we'd have on iOS, right? So yea, maybe if you can explain a little more scenario that will help with figuring out a reasonable set of props across platforms.

CedricGuillemet:
The scenario is quite simple actually: when isTransparent is true for the EngineView, there is nothing to do on iOS or Windows except.
On Android, there are 3 modes: textureView, surfaceView with zTop, surfaceView with with mediaOverlay.
So, we can removed `isTopMost` and have an Android enum property 'androidTransparencyMode`.
Or keep `isTopMost` and add another property that will interfer.
Is it too late to remove isTopMost?

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
```diff
@@ -94,30 +95,37 @@ private void setIsTransparentAndIsTopMost(Boolean isTransparent, Boolean isTopMo
...
                 surfaceHolder.setFormat(PixelFormat.TRANSPARENT);
             }
-            if (isTopMost) {
-                // ZOrder is not dynamic before Android 11. Recreate the surfaceView and set order before adding to the parent
-                // https://developer.android.com/reference/android/view/SurfaceView#setZOrderOnTop(boolean)
+            if ((viewMode.equals("") && isTransparent) || viewMode.equals("SurfaceViewZTopMost"))
+            {
                 this.surfaceView.setZOrderOnTop(true);
+            } else if (viewMode.equals("SurfaceViewZMediaOverlay"))
+            {
```
bghgary, line 119:
```suggestion
            } else if (viewMode.equals("SurfaceViewZMediaOverlay")) {
```

bghgary:
It would be good if @ryantrem can review as well before I sign off.

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
```diff
@@ -94,30 +95,37 @@ private void setIsTransparentAndIsTopMost(Boolean isTransparent, Boolean isTopMo
             this.transparentTextureView.setVisibility(View.GONE);
             this.transparentTextureView = null;
         }
-        if (isTransparent && !isTopMost) {
+
+        //if (isTransparent) {
```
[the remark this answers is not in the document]
CedricGuillemet: Fixed.

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
CedricGuillemet: Fixed

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
```diff
@@ -33,14 +33,14 @@ public final class EngineView extends FrameLayout implements SurfaceHolder.Callb
 
     private SurfaceView xrSurfaceView;
     private boolean isTransparent = false;
-    private boolean isTopMost = false;
+    private String viewMode = "";
```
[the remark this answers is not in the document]
CedricGuillemet: Fixed

Modules/@babylonjs/react-native/README.md
```diff
@@ -106,3 +98,19 @@ e.g.
 ```
 
 Note: Currently only one `EngineView` can be active at any given time. Multi-view will be supported in a future release.
+
+On Android, specific `androidView` property can help set the type of the view used for rendering. Depending on user needs and performance, refer to this table:
```
ryantrem, line 102: Maybe
```suggestion
The Android specific `androidView` property can help set the type of the view used for rendering. Depending on user needs and performance, refer to the table below. [`TextureView`](https://developer.android.com/reference/android/view/TextureView) can be inserted anywhere in the view hierarchy, but is less efficient. [`SurfaceView`](https://developer.android.com/reference/android/view/SurfaceView) can only be full above or fully below the rest of the UI, but is more efficient.
```

Modules/@babylonjs/react-native/README.md
```diff
@@ -106,3 +98,19 @@ e.g.
...
+| Transparent | androidView value        | Description |
+| ----------- | ------------------------ | ----------- |
+| False       | TextureView              | Opaque TextureView
+| False       | SurfaceView              | Simple surfaceView (default when no `androidView` set with Transparency Off)
+| False       | SurfaceViewZTopMost      | SurfaceView with ZTopMost set to true. https://developer.android.com/reference/android/view/SurfaceView#setZOrderOnTop(boolean)
+| False       | SurfaceViewZMediaOverlay | SurfaceView with ZMediaOverlay set to true. https://developer.android.com/reference/android/view/SurfaceView#setZOrderMediaOverlay(boolean)
+| True        | TextureView              | Transparent TextureView, less performant than transparent SurfaceViews
+| True        | SurfaceView              | SurfaceView will stay opaque
+| True        | SurfaceViewZTopMost      | SurfaceView with ZTopMost set to true. (default when no `androidView` set with Transparency On) Transparent but top most.
+| True        | SurfaceViewZMediaOverlay | SurfaceView with ZMediaOverlay set to true. Only Transparent ontop of other SurfaceViews
```
ryantrem, line 113:
```suggestion
| isTransparent | androidView        | Description |
| ----------- | ------------------------ | ----------- |
| False       | TextureView              | Opaque TextureView.
| False       | SurfaceView              | Simple surfaceView (default when no `androidView` set with `isTransparent=false`).
| False       | SurfaceViewZTopMost      | SurfaceView with [ZTopMost](https://developer.android.com/reference/android/view/SurfaceView#setZOrderOnTop(boolean)) set to `true`.
| False       | SurfaceViewZMediaOverlay | SurfaceView with [ZMediaOverlay](https://developer.android.com/reference/android/view/SurfaceView#setZOrderMediaOverlay(boolean)) set to `true`.
| True        | TextureView              | Transparent TextureView.
| True        | SurfaceView              | SurfaceView will stay opaque
| True        | SurfaceViewZTopMost      | SurfaceView with [ZTopMost](https://developer.android.com/reference/android/view/SurfaceView#setZOrderOnTop(boolean)) set to `true`. Transparent but top most. (default when no `androidView` set with `isTransparent=true`)
| True        | SurfaceViewZMediaOverlay | SurfaceView with [ZMediaOverlay](https://developer.android.com/reference/android/view/SurfaceView#setZOrderMediaOverlay(boolean)) set to `true`. Only Transparent on top of other SurfaceViews.
```

Modules/@babylonjs/react-native/EngineView.tsx
```diff
@@ -116,7 +116,7 @@ export const EngineView: FunctionComponent<EngineViewProps> = (props: EngineView
     if (initialized !== false) {
         return (
             <View style={[{ flex: 1 }, props.style, { overflow: "hidden" }]}>
-                { initialized && <NativeEngineView ref={engineViewRef} style={{ flex: 1 }} onSnapshotDataReturned={snapshotDataReturnedHandler} isTransparent={isTransparent} antiAliasing={antiAliasing} isTopMost={isTopMost}/> }
+                { initialized && <NativeEngineView ref={engineViewRef} style={{ flex: 1 }} onSnapshotDataReturned={snapshotDataReturnedHandler} isTransparent={isTransparent} antiAliasing={antiAliasing} androidView={androidView}/> }
```
ryantrem, line 119: If the `androidView` property is set on a control that doesn't support the property (e.g. the underlying component for all platforms other than Android), does it no produce an error or warning or anything?

CedricGuillemet: I didn't see any error or warning.

bghgary (approved):
Good from my perspective.

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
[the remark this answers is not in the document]
ryantrem: I assume the "remove" comment was about the commented out `//if (isTransparent) {`, which appears to still be there.

ryantrem (approved):

Modules/@babylonjs/react-native-iosandroid/android/src/main/java/com/babylonreactnative/EngineView.java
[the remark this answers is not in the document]
CedricGuillemet: Indeed! idk what happened. Gone now.

PR #463, opened 2022-09-29
Use abort controller to prevent async setState after unmount.

nima-ap:
**Describe the change**

Only calls setState if not unmounted. useModuleInitializer was trying to set state in an async fashion, and this must always be accompanied by some cancellation or aborting mechanism in React, otherwise it is possible for a state update to happen on an unmounted component, causing memory leaks and warnings.

**Screenshots**

N/A

**Documentation**

N/A

**Testing**

*I need help testing on [Android|iOS|Windows]*


ryantrem:
I think technically in this case it is [safe to ignore the warning](https://github.com/reactwg/react-18/discussions/82) (which as been removed in newer versions of React Native!), but the warning is also noisy and distracting so avoiding it still seems like a good idea to me.

ryantrem (approved):

PR #472, opened 2022-10-07
fix(NativeEngineHook): avoid deprecated warning

RodSarhan:
Fixed useAppState hook causing "removeEventListener" deprecated warning on newer react native versions

**Describe the change**

- Checks if the return of AppState.addEventListener has a remove method
- If the remove method exists it will be called
- if the remove method doesn't exist, the event listener will be removed with the deprecated removeEventListener method

**Documentation**

*N/A*

**Testing**

Tested on Android and iOS

### DIFF
--- a/Modules/@babylonjs/react-native/NativeEngineHook.ts
+++ b/Modules/@babylonjs/react-native/NativeEngineHook.ts
@@ -34,11 +34,16 @@ function useAppState(): string {
             setAppState(appState);
         };
 
-        AppState.addEventListener("change", onAppStateChanged);
+        const appStateListener = AppState.addEventListener("change", onAppStateChanged);
 
         return () => {
-            AppState.removeEventListener("change", onAppStateChanged);
-        }
+            if (!!appStateListener?.remove) {
+                appStateListener.remove();
+            }
+            else {
+                AppState.removeEventListener("change", onAppStateChanged);
+            }
+        };
     }, []);
 
     return appState;


ryantrem (approved):

ryantrem:
Thanks for the contribution @RodSarhan! It looks like this branch is out-of-date with master. Can you merge master, either locally or I think you can directly in the PR as well?

RodSarhan:
@ryantrem It should be good now, thanks for the awesome work you guys are doing!

RodSarhan:
I noticed that TS complained about the method "remove" not existing on appStateListener on react native 64
We can avoid that by asserting the type, please let me know if you prefer that

```
const appStateListener = AppState.addEventListener("change", onAppStateChanged);
// Asserting the type to prevent TS type errors on older RN versions
const removeListener = appStateListener?.["remove"] as undefined | Function;

return () => {
        if (removeListener) {
                removeListener();
        }
        else {
                AppState.removeEventListener("change", onAppStateChanged);
        }
};
```

CedricGuillemet:
@ryantrem any opinion on this one?

ryantrem:
Oy did this PR merge even though there was a TS build error? I wonder if we have our checks misconfigured...
I think the suggested change above seems reasonable. If it fixes a TS build break, then it'd be great if we could get the change soon, and then we can always consider alternatives in the future.

RodSarhan:
#475 should fix this

PR #473, opened 2022-10-07
Update BabylonNative and Babylon.js

CedricGuillemet:
Tested RN 0.65 on Windows, iOS, Android


ryantrem:
Won't this break Playground on Android since it doesn't have the XHR header fix yet?

bghgary:
> Won't this break Playground on Android since it doesn't have the XHR header fix yet?

It should be fine. We are returning null for getResponseHeader on other platforms which should make it not throw an exception anymore. It still won't work if someone depends on the result, but that's not the case here.

bghgary (approved):

PR #484, opened 2022-10-13
Revoke rendering when SCP is unloaded

CedricGuillemet:
When the view is unloaded, rendering keeps going and initialization promise created by resetview call is resolve almost immediately. Initialization is not done for a second view and no rendering is visible.

### DIFF
--- a/Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
+++ b/Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
@@ -19,6 +19,13 @@ namespace winrt::BabylonReactNative::implementation {
 
         _revokerData.SizeChangedRevoker = SizeChanged(winrt::auto_revoke, { this, &EngineView::OnSizeChanged });
 
+        // Stop rendering when SwapChainPanel is unloaded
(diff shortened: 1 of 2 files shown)


Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/EngineView.cpp
```diff
@@ -19,6 +19,13 @@ namespace winrt::BabylonReactNative::implementation {
 
         _revokerData.SizeChangedRevoker = SizeChanged(winrt::auto_revoke, { this, &EngineView::OnSizeChanged });
 
+        // Stop rendering when SwapChainPanel is unloaded
+        _revokerData.UnloadedEventToken = Unloaded(winrt::auto_revoke, [ref = get_weak()](auto const&, auto const&) {
+            if (auto self = ref.get()) {
+                self->_revokerData.RenderingRevoker.revoke();
```
ryantrem, line 25: This is asymmetrical and won't work if the control is added/removed to/from the xaml tree multiple times. Loaded happens when the control is added to the xaml tree, Unloaded happens when the control is removed from the xaml tree. Given this, I think it makes sense to only register Loaded and Unloaded handlers in the constructor, and then do the rest of the init/deinit in the Loaded/Unloaded handlers. This would make this work correctly with multiple iterations of adding/removing the control to/from the xaml tree.

CedricGuillemet:
No the right solution to the problem. Closing.

PR #494, opened 2022-10-19
RN Windows/extra rendering frame bug

CedricGuillemet:
With RN Windows, when the view was unmounted, rendering calls keep being performed and view is not destroyed.
Added mount/unmount event to handle rendering.
With this fix in place, because of async, an extra frame rendering might happen after the resetView. As this resolves the initialization promise, its creation is done when a new engine is being created.

### DIFF
--- a/Modules/@babylonjs/react-native/shared/BabylonNative.cpp
+++ b/Modules/@babylonjs/react-native/shared/BabylonNative.cpp
@@ -115,7 +115,7 @@ namespace BabylonNative
                 });
             }
             m_isRenderingEnabled = true;
-            m_pendingReset = false;
+            m_newEngine = false;
         }
 
         void UpdateMSAA(uint8_t value)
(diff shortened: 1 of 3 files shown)


Modules/@babylonjs/react-native/shared/BabylonNative.cpp
```diff
@@ -158,13 +153,17 @@ namespace BabylonNative
             }
         }
 
+        void newEngine()
```
bghgary, line 156:
```suggestion
        void NewEngine()
```

But it's still kind of a weird name. Maybe `Initialize()`?

bghgary (approved):

Modules/@babylonjs/react-native/shared/BabylonNative.cpp
CedricGuillemet: I agree, the name is weird. Initialize is too generic and I'd like to send the message that this function is called when a new engine is created. `EngineInitializationPending` ?

bghgary: I don't think it's too generic. It's called from the static Initialize function. The code happens to be setting a flag for engine, but it could potentially do other things.

PR #508, opened 2022-12-05
Set Windows SDK Version in build scripts

CedricGuillemet:
Build Script var to set Windows SDK version

### DIFF
--- a/Package/gulpfile.js
+++ b/Package/gulpfile.js
@@ -124,7 +124,17 @@ const initializeSubmodulesMostRecentBabylonNative = async () => {
 
 const makeUWPProjectPlatform = async (name, arch) => {
   shelljs.mkdir('-p', `./../Modules/@babylonjs/react-native/Build/uwp_${name}`);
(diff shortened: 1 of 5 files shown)


Package/gulpfile.js
```diff
@@ -124,7 +124,17 @@ const initializeSubmodulesMostRecentBabylonNative = async () => {
...
+
+  let WindowsSDKVersionCMakeString = "";
+  const [WindowsSDKVersionFlag] = process.argv.slice(3);
+  if (WindowsSDKVersionFlag === '-WindowsSDKVersion') {
+    const [WindowsSDKVersionValue] = process.argv.slice(4);
+    if (WindowsSDKVersionValue)
+    {
+      WindowsSDKVersionCMakeString = `-DCMAKE_SYSTEM_VERSION=${WindowsSDKVersionValue}`;
+    }
+  }
```
ryantrem, line 136: Would it make more sense to put this logic in tools.js (which I just noticed we have duplicated between RN versions 😬) and invoked via an npm script in package.json? This would be more like what we do for iOS, which makes it really easy to do locally in an isolated way.

CedricGuillemet: I'm not sure. tools.js is used to build the PG. and gulp.js is used to build packages. moreover, you can test this change locally with `npx gulp buildUWPPublish -WindowsSDKVersion 10.0....`

CedricGuillemet:
fixed with https://github.com/BabylonJS/BabylonReactNative/pull/665

PR #511, opened 2022-12-07
React-native 0.70 support

CedricGuillemet:
React-native 0.70 package as 0.69 is not compatible.

Tested with partners, on Android.


Apps/Playground/0.70/metro.config.js
```diff
@@ -0,0 +1,36 @@
+/**
```
ryantrem, line 1: Don't we need all our customizations to this file, like support for sym linking?

Do we have all the steps for adding a new RN version documented? Seems like it would be good to have a step by step so we can be sure we got everything when we add a new version.

Apps/Playground/0.70/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+// prettier-ignore
```
ryantrem, line 1: Should we use a common tsconfig for all RN versions?

.github/workflows/publish.yml
```diff
@@ -126,6 +148,13 @@ jobs:
         working-directory: ./Package/Assembled-iOSAndroid0.69
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      - name: Version & Publish Package @babylonjs/react-native-iosandroid-0-70
```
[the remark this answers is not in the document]
CedricGuillemet: yes, we talked about it some time ago. I agree it's the thing to do.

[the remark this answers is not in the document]
CedricGuillemet: I think it should be addressed while doing that : https://github.com/BabylonJS/BabylonReactNative/issues/516

Apps/Playground/0.70/tsconfig.json
CedricGuillemet: Might be possible. Not part of this PR.

Apps/Playground/0.70/metro.config.js
ryantrem: It looks like this file still hasn't been updated. Have you run the 0.70 Playground app locally, because it looks like it shouldn't work.

Apps/Playground/0.70/scripts/tools.js
```diff
@@ -0,0 +1,37 @@
+const os = require('os');
```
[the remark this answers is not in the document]
ryantrem: I think worse case you could move all the logic to a common file, pull it into this file via a `require`, and then just delegate to that helper file. Could be done in some other change, just not sure we'll ever get to it if we don't do it with one of these changes to add support for a new RN version. :)

ryantrem (approved):

PR #513, opened 2022-12-07
Update BabylonNative submodule and Babylon.js package version

bghgary:
**Describe the change**

Update Babylon Native submodule and Babylon.js dependency versions. There was a PROTOCOL_VERSION update and a bgfx update.

**Testing**

- [x] Android
- [x] iOS


CoPrez (approved):

CedricGuillemet (approved):
LGTM!

Apps/PackageTest/0.63.1/package.json
```diff
@@ -10,8 +10,8 @@
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx"
   },
   "dependencies": {
-    "@babylonjs/core": "5.6.1",
-    "@babylonjs/react-native": "^0.4.0-alpha.47",
+    "@babylonjs/core": "5.35.1",
+    "@babylonjs/react-native": "^1.4.1",
```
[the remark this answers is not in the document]
ryantrem: I'm guessing this is broken since it doesn't even reference any of the BRN binary packages. But yea removing it or fixing it in a separate PR makes sense.

Apps/Playground/0.65/ios/Playground.xcodeproj/project.pbxproj
```diff
@@ -485,6 +484,7 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
 				CURRENT_PROJECT_VERSION = 1;
+				DEVELOPMENT_TEAM = UBF8T346G9;
```
ryantrem, line 487: I don't think you want to check in these `DEVELOPMENT_TEAM` changes.

Apps/Playground/0.65/npm_run_start_loop.cmd
```diff
@@ -0,0 +1,6 @@
+:Loop
```
ryantrem, line 1: What is this? If we want some tooling like this, we should put it in a shared node script somewhere.

Modules/@babylonjs/react-native-iosandroid/android/build.gradle
```diff
@@ -94,7 +94,6 @@ android {
     }
     externalNativeBuild {
         cmake {
-            version '3.19.6'
```
ryantrem, line 97: Why this change?

Apps/Playground/0.65/npm_run_start_loop.cmd
bghgary: Oops, that's not supposed to be in the PR. :)

Apps/Playground/0.65/ios/Playground.xcodeproj/project.pbxproj
bghgary: Well, that's annoying. We have to undo the dev team before merging? I've never had to do that. I wonder what's different?

Modules/@babylonjs/react-native-iosandroid/android/build.gradle
bghgary: Ugh, sorry, this isn't supposed to be there either. I run with a newer version of cmake.

Apps/Playground/0.65/ios/Playground.xcodeproj/project.pbxproj
bghgary: Weird, I don't know what happened, but it didn't add them this time. I was having trouble with setting the dev team, so maybe I touched something that caused xcode to write it out. Anyways, it's removed now.

Apps/PackageTest/0.63.1/package.json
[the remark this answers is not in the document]
bghgary: We can do it in a separate PR.

PR #541, opened 2023-01-17
Updated Babylon Native version

SergioRZMasson:
Updated Babylon Native to get the update to viewport assignment, allowing for the usage of multiple cameras rendering to the same render target. 

For more information: https://github.com/BabylonJS/BabylonNative/pull/1183


bghgary:
Nice, I need to do this also for the URL processing fixes.

bghgary (approved):

PR #569, opened 2023-03-13
test BN protocol tool and github action

CedricGuillemet:
Runs a list of BabylonNative commit against a range of Babylon.js NPM version to list compatible versions.
The list is then uploaded as an artifact.

### DIFF
--- a/.github/workflows/test_protocol.yml
+++ b/.github/workflows/test_protocol.yml
@@ -1,0 +1,29 @@
+name: Test BabylonNative Protocol compatibility
+on:
+  workflow_dispatch:
+    inputs:
+      NPM_Mask:
+        description: 'NPM versions mask'
+        required: true 
+        type: string
+        default: 5.*.*
+
+jobs:
+  build:
+    runs-on: windows-2019
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@master
+    - name: NPM Install
+      run: npm install
+      working-directory: ./Apps/TestProtocol
+    - name: Run tests
+      run: npm run testPackages ${{ github.event.inputs.NPM_Mask }}
+      working-directory: ./Apps/TestProtocol
+    - name: Upload Result 
+      uses: actions/upload-artifact@v2
+      with:
+          name: NPMCompatibility
+          path: |
+            ./compatibility.json
--- a/Apps/TestProtocol/testBNProtocol.js
+++ b/Apps/TestProtocol/testBNProtocol.js
@@ -1,0 +1,130 @@
+const fs = require('fs');
+const shelljs = require('shelljs');
+
+let compatibility = [];
+function execute(command, workingDirectory, callback){
+    const res = shelljs.exec(command, { fatal: false, cwd: workingDirectory, async: false });
+    callback(res.code, res.stdout, res.stderr);
+};
+
+function copy(source, destination) {
+    fs.copyFileSync(source, destination);
+    console.log(`${source} was copied to ${destination}`);
+}
+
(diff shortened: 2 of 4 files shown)


Apps/TestProtocol/testBNProtocol.js
```diff
@@ -0,0 +1,130 @@
...
+                    }
+                });
+            });
+        });
+    });
+    console.log(`Compatible versions for BabylonReactNative ${tag} using BabylonNative ${hash}:`, compatiblePackageVersions);
+    compatibility.push({tag:tag, hash:hash, npms:compatiblePackageVersions});
+}
+
+BRNVersions.forEach((versionToTest) =>{
```
ryantrem, line 125: Is it a goal to be able to run this for every version locally? Would it make more sense for the version to be an input to the script, and then use a GitHub actions matrix thing to be able to run them all in parallel across multiple machines?

Apps/TestProtocol/testBNProtocol.js
```diff
@@ -0,0 +1,130 @@
...
+        // already patched
+        return;
+    }
+    const newContent = protocolChecker + testScript;
+    fs.writeFileSync(filePath, newContent);
+}
+
+function checkoutAndBuildBN(tag, hash, callback) {
+    console.log("Git clone.");
+    execute('git clone https://github.com/BabylonJS/BabylonNative.git', './', (error, stdout, stderr) => {
```
ryantrem, line 52: Oy. Can we use Promises and async/await in this code? It's just a node.js script, right?

Apps/TestProtocol/testBNProtocol.js
CedricGuillemet: Yes, it would be faster to run. And it should also use the result from last run to not re-test everything.

Apps/TestProtocol/testBNProtocol.js
CedricGuillemet: Yes, it's a Node.js script.

.github/workflows/test_protocol.yml
```diff
@@ -0,0 +1,29 @@
+name: Test BabylonNative Protocol compatibility
```
bghgary, line 1: Like we discussed offline, this isn't just testing the protocol version. Rename folder to TestCompatibility, etc.?

Documentation/TestingCompatibility.md
```diff
@@ -0,0 +1,35 @@
+# Testing BabylonNative Compatibility
+
+`Apps/TestCompatibility` is a semi-automated tool to test compatibility between BabylonReactNative (more specificaly BabylonNative)
+and Babylon.js NPM packages.
+It checkouts, build and run validation tests for every BabylonNative hash version from a list and every Babylon.js NPM package.
```
docEdub, line 5:
```suggestion
It checks out, builds and runs validation tests for every BabylonNative hash version from a list and every Babylon.js NPM package.
```

docEdub (approved):

Apps/TestCompatibility/testBNCompatibility.js
```diff
@@ -0,0 +1,134 @@
...
+        // already patched
+        return;
+    }
+    const newContent = protocolChecker + testScript;
+    fs.writeFileSync(filePath, newContent);
+}
+
+function checkoutAndBuildBN(tag, hash, callback) {
+    console.log("Git clone.");
+    execute('git clone https://github.com/BabylonJS/BabylonNative.git', './', (error, stdout, stderr) => {
```
ryantrem, line 49: Can we use async/await here to make this more readable?

PR #577, opened 2023-05-11
fix gradle and cmake for windows local PG 0.71 builds

CedricGuillemet:


Modules/@babylonjs/react-native-iosandroid/android/CMakeLists.txt
```diff
@@ -68,9 +68,11 @@ if (EXISTS "${TURBOMODULE_DIR}/CMakeLists.txt")
         set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ccache)
     endif(CCACHE_FOUND)
 
-    set(BUILD_DIR ${PROJECT_BUILD_DIR})
-    if(CMAKE_HOST_WIN32)
-            string(REPLACE "\\" "/" BUILD_DIR ${BUILD_DIR})
+    if (NOT "${PROJECT_BUILD_DIR}" STREQUAL "")
```
ryantrem, line 71: Can you add a comment? It's not clear to me why this is required now but didn't used to be.

Modules/@babylonjs/react-native-iosandroid/android/CMakeLists.txt
```diff
@@ -68,9 +68,11 @@ if (EXISTS "${TURBOMODULE_DIR}/CMakeLists.txt")
...
+        set(BUILD_DIR ${PROJECT_BUILD_DIR})
+        if(CMAKE_HOST_WIN32)
+                string(REPLACE "\\" "/" BUILD_DIR ${BUILD_DIR})
```
ryantrem, line 74:
```suggestion
            string(REPLACE "\\" "/" BUILD_DIR ${BUILD_DIR})
```

Apps/Playground/0.71/android/app/build.gradle
```diff
@@ -102,6 +102,28 @@ android {
         versionName "1.0"
     }
 
+    packagingOptions {
+        pickFirst 'lib/*/libfabricjni.so'
```
ryantrem, line 106: Isn't `pickFirst` typically needed when there are duplicate libs? If that's the case, why are there duplicates of all these? Or is something else going on? Maybe add some comments here as well.

CedricGuillemet: This is the error I get when building a freshly created playground with rn0.71 & Babylon 
image

CedricGuillemet: too many .so files coming from react-native were copied into assembled and into the npm. jnilibs is now more restrictive on files.

ryantrem (approved):

PR #589, opened 2023-07-18
Update BN submodule

CedricGuillemet:
Update BN to have this particular commit in : https://github.com/BabylonJS/BabylonNative/commit/c89031ec58bca5faffa92d3498230c5484175e57

### DIFF
--- a/.github/workflows/ios_android.yml
+++ b/.github/workflows/ios_android.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Setup CMake
         uses: jwlawson/actions-setup-cmake@v1.8
         with:
-          cmake-version: '3.19.6' # See https://gitlab.kitware.com/cmake/cmake/-/issues/22021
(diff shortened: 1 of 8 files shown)


.github/workflows/ios_android.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       - name: Setup CMake
         uses: jwlawson/actions-setup-cmake@v1.8
         with:
-          cmake-version: '3.19.6' # See https://gitlab.kitware.com/cmake/cmake/-/issues/22021
+          cmake-version: '3.26.3' # See https://gitlab.kitware.com/cmake/cmake/-/issues/22021
```
ryantrem, line 31: Remove the comment with the link to the bug on all these.

ryantrem (commented):
When I recently tested, cmake 3.25+ didn't work with BRN for iOS because of a cmake change with location of built archive (static lib) files, which resulted in linker errors. @bghgary  has more details on this.

bghgary (approved):

PR #598, opened 2023-09-07
Basekit

CedricGuillemet:
Some partners do not use XR/Camera but because it's in the binary, apps needs permission. Moreover, asking for permission while not using the feature is sufficient to be rejected by the store.
We need to provide a way to not use these feature. Doing it at realtime means permission is still needed and, I guess, a binary import check is also performed when uploading to the store.
The approach here is to release a new package per platform/RN version, suffixed with `basekit`.
The equivalent basekit of `@babylonjs/react-native-windows-0-71` is `@babylonjs/react-native-basekit-windows-0-71`
In order to not explode build times (they are already way too big), per platform/RN version, build steps are added in gulp instead of adding new jobs.
Basically, BRN is built as before, then, some flags are changed and a private preprocessor is added to BabylonReactNative project to not compile with XR/NativeCamera. Another build happens and binaries are copied to a different Assembled directory.
Then, 2 artifacts and packages are published.

Note: Plan for the future is to remove the legacy NPM packages and replace them with BaseKit ones + 1 package per feature (XR, Camera,...) 

Both BaseKit and 'FullKit' contain same files. NativeCamera and XR libraries/headers will be present in the NPM. This is for ease of development/maintenance. Difference is only done at BabylonNative compilation and therefore, at link for any users.

There is a lot of copy/paste in the GitHub Action jobs that will be addressed in a following PR.

### DIFF
--- a/Package/gulpfile.js
+++ b/Package/gulpfile.js
@@ -683,10 +683,8 @@ const patchPackageVersion = async () => {
 const copyFiles = gulp.parallel(copyIOSAndroidCommonFiles, copyIOSFiles, copyAndroidFiles);
 
 const buildTS = gulp.series(patchPackageVersion, copyCommonFiles, copySharedFiles, buildTypeScript, validateAssembled);
-const build = gulp.series(patchPackageVersion, buildIOS, buildAndroid, createIOSUniversalLibs, copyFiles, validateAssemblediOSAndroid,
-  switchToBaseKit,
-  patchPackageVersion, buildIOS, buildAndroid, createIOSUniversalLibs, copyFiles, validateAssemblediOSAndroid
-  );
+const buildIOSAndroid = gulp.series(patchPackageVersion, buildIOS, buildAndroid, createIOSUniversalLibs, copyFiles, validateAssemblediOSAndroid);
+const build = gulp.series(buildIOSAndroid, switchToBaseKit, buildIOSAndroid);
 const rebuild = gulp.series(clean, build);
 const pack = gulp.series(rebuild, createPackage);
 


Package/gulpfile.js
```diff
@@ -654,7 +683,10 @@ const patchPackageVersion = async () => {
 const copyFiles = gulp.parallel(copyIOSAndroidCommonFiles, copyIOSFiles, copyAndroidFiles);
 
 const buildTS = gulp.series(patchPackageVersion, copyCommonFiles, copySharedFiles, buildTypeScript, validateAssembled);
-const build = gulp.series(patchPackageVersion, buildIOS, buildAndroid, createIOSUniversalLibs, copyFiles, validateAssemblediOSAndroid);
+const build = gulp.series(patchPackageVersion, buildIOS, buildAndroid, createIOSUniversalLibs, copyFiles, validateAssemblediOSAndroid,
+  switchToBaseKit,
+  patchPackageVersion, buildIOS, buildAndroid, createIOSUniversalLibs, copyFiles, validateAssemblediOSAndroid
+  );
```
ryantrem, line 689: Maybe remove the duplication of core build steps?
```suggestion
const buildIOSAndroid = gulp.series(patchPackageVersion, buildIOS, buildAndroid, createIOSUniversalLibs, copyFiles, validateAssemblediOSAndroid);
const build = gulp.series(buildIOSAndroid, switchToBaseKit, buildIOSAndroid);
```

ryantrem:
I know you mentioned deduping copy/paste in GitHub Actions, I imagine we could also remove some of the duplication in gradle files, pod files, etc. as well. Not sure how much effort is worth putting into that if there is a broader plan for better modularization, but also not sure when that will happen, so hopefully in the meantime we keep the maintenance burden in check.

ryantrem (approved):

PR #606, opened 2023-09-27
fix permissions after basekit changes

CedricGuillemet:
Fix the regressions appearing after https://github.com/BabylonJS/BabylonReactNative/pull/598
An error with RNPermissionsModule not found.
I've reverted the change in the json and everything is fine again. 
Why is dependency needed in react-native package as it's already present in other packages like https://github.com/BabylonJS/BabylonReactNative/blob/313ee5c48003f0be1b4e6231a0cfea8aafec9155/Modules/%40babylonjs/react-native-iosandroid/package.json#L34 ?
Can this be an issue for basekit package ? @nima-ap

### DIFF
--- a/Modules/@babylonjs/react-native/package.json
+++ b/Modules/@babylonjs/react-native/package.json
@@ -30,7 +30,8 @@
   "peerDependencies": {
     "@babylonjs/core": "^6.14.0",
     "react": ">=16.13.1",
-    "react-native": ">=0.63.1"
+    "react-native": ">=0.63.1",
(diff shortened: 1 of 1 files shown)


ryantrem:
Strange. I think this must be the rn permissions js code is looking for a file by relative path or something. This probably doesn't repro with the published packages, it's probably a side effect of the local packages used within the BRN repository. That said, I also doubt this reference will cause a problem for those who are trying to avoid using native APIs that require permissions.

PR #608, opened 2023-09-29
Update to Babylon Native master and fix many issues

ryantrem:
**Describe the change**

This change updates to the Babylon Native master branch, and fixes many issues (build regressions, runtime regressions, and issues due to newer tooling).

- Since dependencies are no longer submodules, but rather dynamically downloaded into the _deps folder, and since there is one _deps folder per platform and architecture, and since CMake was outputting to a source directory, we ended up in a situation where metro was picking up "duplicate" package.json files from all the _deps and erroring out. We don't want Metro picking up those package.json files anyway, and CMake should be outputting to a common Build directory, so it's been reconfigured to output to a Build directory from the repo root. This also required some other updates, like the xcworkspace files pulling in the generated Xcode project from a different folder.
- There were build breaks in most versions of React Native with a recentish Xcode update. These were fixed in minor updates to the React Native releases, so I updated both 0.69 and 0.70 to slightly newer releases. See https://github.com/facebook/react-native/issues/34106#issuecomment-1495970568
- The podspecs were picking up ios/**/*.{h,m,mm}, which prior to the fix in bullet 1 were picking up tons of files from _deps. I fixed this by removing the ** before fixing bullet 1, but I'm going to leave it like this as it is safer anyway.
- Napi::Attach is no longer templated, so fixed this.
- glslang-default-resource-limits was an added dependency in BN, so it needed to be added here as well. 
- Updated gradle packagingOptions to exclude more files resulting in build breaks on Android (thanks @bghgary!).
- Fixed an issue where fbJni installation at gradle sync time was failing (thanks @bghgary!).
- Configure CMake iOS tool chain via cmake fetch since it is no longer a submodule.
- Pulled in several fixes made to other repos along the way:
  - https://github.com/BabylonJS/JsRuntimeHost/commit/5fefa63d9af36df062ef7cb275ce97cac506057e
  - https://github.com/BabylonJS/JsRuntimeHost/commit/b3d4f2cb8c86fb55dc4e8ae78ba5f5a98a5be6ba
  - https://github.com/BabylonJS/BabylonNative/commit/71a89f75bf733debc112eefd7b5603cb0f9bf63a
- Added _deps link folders in Windows vcxproj and removed deprecated paths.
- Disabled UnityBuild for UrlLib because of a conflict of headers.
- Added glslang-default-resource-limit link library.
- Explicitly set CMake Windows projects output to be VS2019 or CMake will use VS2022 when available and by default.
- Disable warning in NativeCamera for unreachable code BabylonNative PR : https://github.com/BabylonJS/BabylonNative/pull/1302

**Screenshots**

N/A

**Documentation**

N/A

**Testing**

I've tested Android debug and iOS debug. Clang segfaults on my machine when compiling NativeInput.cpp in release in Xcode, so I could not test iOS release. Android crashes on launch in release, but it is very early in React Native specific code, so I don't think this is a result of updating BN. I did not test Windows at all.


Package/gulpfile.js
```diff
@@ -75,7 +75,7 @@ const buildTypeScript = async () => {
 
 const makeXCodeProj = async () => {
   shelljs.mkdir('-p', 'iOS/Build');
-  exec(`cmake -G Xcode -DCMAKE_TOOLCHAIN_FILE=../../../Apps/Playground/Playground/node_modules/@babylonjs/react-native-iosandroid/submodules/BabylonNative/Dependencies/ios-cmake/ios.toolchain.cmake -DPLATFORM=OS64COMBINED -DENABLE_ARC=0 -DENABLE_BITCODE=1 -DDEPLOYMENT_TARGET=12 -DENABLE_PCH=OFF ${cmakeBasekitBuildDefinition} ..`, 'iOS/Build');
+  exec(`cmake -B Build -G Xcode ${cmakeBasekitBuildDefinition}`, 'iOS');
```
bghgary, line 78: This seems like it's missing `-D IOS=ON`. How is it working without it?

ryantrem: The IOS flag affects BN's top level CMakeLists.txt, but that is not the top level CMakeLists.txt for BRN. The top level ones for BRN are `Modules/@babylonjs/react-native-iosandroid/ios/CMakeLists.txt` and `Package/iOS/CMakeLists.txt`, both of which set the same cmake variables that BN's CMakeLists.txt conditionally does based on the IOS flag.

bghgary (approved):

PR #610, opened 2023-10-04
rename and set node 16

CedricGuillemet:
- Set Node16 for test package GitHub Action
- Rename .yml to be compliant with others
- Set Mac OS 12 explicitly so xcode15 will not be used by default


bghgary (approved):
I think we have flipflopped between macos-latest and a specific version a few times. Not sure which is best.

PR #611, opened 2023-10-05
Exclude XR and camera libs from basekit podspec

CedricGuillemet:
XR, nativeXR and nativeCamera libraries are not excluded from BaseKit podspec resulting in missing symbols at link times.
List libraries explicitely to fix the issue.

### DIFF
--- a/Package/BaseKit/react-native-babylon.podspec
+++ b/Package/BaseKit/react-native-babylon.podspec
@@ -16,10 +16,43 @@ Pod::Spec.new do |s|
   s.source_files = "ios/**/*.{h,m,mm}"
   s.requires_arc = true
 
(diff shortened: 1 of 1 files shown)


ryantrem:
Shouldn't the files themselves (e.g. libxr.a) be excluded from the package? Then the podspec doesn't need to change.

Alternatively, if we must include unused libs in the package for some reason, is there some kind of podspec exclusion syntax, so we can still include *.a but then exclude specific libs?

CedricGuillemet:
excluding .a from the package needs some work in the gulp file and it's starting to be (a little) convoluted and hard to maintain. I don't want to add cases where files are deleted at some steps.

In the pod spec, I tried to use `exclude_files` without success. It seems only valid for source files.

ryantrem (approved):

PR #613, opened 2023-10-17
CMake Fetchcontent instead of submodules

CedricGuillemet:
To be more consistent with BN, easier updates to different BabylonNative version and prepare the work to use prebuilt BN.

- fix issue with arcore. "windows build agents don't support the path lengths required for initializing arcore dependencies". module is not fetched anymore on Windows.
- Remove code to update submodule in `gulp.js` , `tools.js` andjobs .yml
- Code to update submodule to the latest in `gulp.js` is removed. Now change the commit id in `shared/CMakeLists.txt`

Notes:
- Using fetch populate and add_subdirectory after instead of makeAvailable because of binary directory differences of platforms. This is something I'll try to revisit in upcoming PRs.
- There is a check to not populate multiple times in  `shared/CMakeLists.txt`.  With iOS build, I got an error saying it's already populated. I'm sure why this happens.

This change needs https://github.com/BabylonJS/BabylonNative/pull/1312


bghgary (approved):
Looks okay to me

PR #615, opened 2023-10-20
BabylonReactNative Basekit frontend package

CedricGuillemet:
Front end package named `@babylonjs/react-native-basekit` that downloads and installs necessary component for iOS/Android or Windows. Same as 'legacy' basekit (no XR or Camera).

- [x] exception/error handling: I can see errors reported by the package when doing npm install with PG 
- [x] env vars for URL sources
- [x] Cache mecanism
- [x] Test with playground test app iOS/Android/Windows
- [x] forward release from github release in publish.yml
- [x] npm install react-native@latest -> does it re-run package post install? Yes it does!
- [x] npm install BRN then install react-native-windows -> is BRN package updated as well ? No, I had to run `npm ci` Is it normal usage ?
- [x] readme updated without mention of permissions. Also removed package name depending on react-native version


Modules/@babylonjs/react-native-basekit/scripts/install.js
```diff
@@ -0,0 +1,174 @@
+const https = require('https');
+const zlib = require('zlib');
+const tar = require('tar');
+const path = require('path');
+const fs = require('fs');
+
+function getCachePathCandidates() {
+  return [
+    process.env.npm_config_sass_binary_cache,
```
RaananW, line 9: probably a leftover?

Modules/@babylonjs/react-native-basekit/scripts/install.js
```diff
@@ -0,0 +1,174 @@
...
+
+  return flags[index + 1];
+}
+
+function getBinaryUrl(package, binaryFilename) {
+  var site = getArgument('--brn-binary-site') ||
+              process.env.BRN_BINARY_SITE  ||
+              process.env.npm_config_brn_binary_site ||
+              (package.nodeBabylonConfig && package.nodeBabylonConfig.binarySite) ||
+              'https://github.com/CedricGuillemet/BabylonReactNative-1/releases/download';
```
RaananW, line 55: should this stay on github? and if yes - should it stay on your github (and not babylon)?

Modules/@babylonjs/react-native-basekit/scripts/install.js
```diff
@@ -0,0 +1,174 @@
...
+    if ([301, 302, 303, 307, 308].includes(response.statusCode)) {
+      // If the response is a redirect, recursively call downloadExtractAndCache with the new URL
+      downloadExtractAndCache(response.headers.location, cachedFilePath);
+    } else if (response.statusCode === 200) {
+      const gunzip = zlib.createGunzip();
+      const untar = tar.extract({ cwd: process.cwd() }); // Extract to the current working directory
+
+      // create cache directory to store download file
+      try {
+        fs.mkdirSync(path.dirname(cachedFilePath), {recursive: true});
```
RaananW, line 75: mkdir (and mkdirSync) could throw an error if the directory already exists. You can check if it exists before. something like this:
```
if (!fs.existsSync(dir)){
    fs.mkdirSync(dir, {recursive: true});
}
```

Modules/@babylonjs/react-native-basekit/scripts/install.js
```diff
@@ -0,0 +1,174 @@
...
+      })
+      .on('end', (err) => {
+        const fileData = Buffer.concat(chunks);
+        fs.writeFile(cachedFilePath, fileData, (err) => {
+          if (err) {
+            console.error('Error writing BabylonReactNative cache file:', err);
+            process.exit(1);
+          }
+        });
+      })
```
RaananW, line 93: unless you need the chunks you could create a writeStream and pipe the response to it instead:

https://stackoverflow.com/a/22907134/729146

Modules/@babylonjs/react-native-basekit/scripts/install.js
```diff
@@ -0,0 +1,174 @@
...
+      .pipe(gunzip) // Unzip the response
+      .pipe(untar)  // Extract the tar file
```
RaananW, line 95: just wondering if it is the same process for windows binaries?

Modules/@babylonjs/react-native-basekit/scripts/install.js
```diff
@@ -0,0 +1,174 @@
...
+  if (projectPackageJson.dependencies && projectPackageJson.dependencies['react-native']) {
+    const reactNativeVersion = projectPackageJson.dependencies['react-native'];
+    console.log(`Detected React Native version: ${reactNativeVersion}`);
+    reactNativePostfix = "";
+    if (reactNativeVersion >= '0.71') {
+      reactNativePostfix = "0.71";
+    } else if (reactNativeVersion >= '0.70') {
+      reactNativePostfix = "0.70";
+    } else if (reactNativeVersion >= '0.69') {
+      reactNativePostfix = "0.69";
```
RaananW, line 134: do you mean to compare strings here? I mean, it's javascript and it will (probably as expected) work because of coercion,  but using numbers directly would make more sense

Modules/@babylonjs/react-native-basekit/scripts/install.js
```diff
@@ -0,0 +1,174 @@
...
+    if (reactNativeWindowsVersion) {
+      console.log("react-native-windows detected.");
+      if (!useCachedFile(windowsArchiveCachedPath)) {
+        const reactNativeiOSWindows = getBinaryUrl(packageJson, windowsArchive);
+        downloadExtractAndCache(reactNativeiOSWindows, windowsArchiveCachedPath);
+      }
+    }
+  } else {
+    console.error("No react-native version found for BabylonReactNative.");
+    process.exit(1);
```
RaananW, line 170: general idea (totally not a must) if you want to catch the exact error (for example during CI testing), give them different code values

Modules/@babylonjs/react-native-basekit/scripts/install.js
```diff
@@ -0,0 +1,174 @@
...
+    const sourceStream = fs.createReadStream(filePath); 
+    sourceStream
+    .pipe(gunzip) // Decompress the .tar.gz file
+    .pipe(untar) // Extract the tar contents to the cwd
+    .on('error', (err) => {
+      console.error('Error extracting .tar.gz file:', err);
+      process.exit(1);
+    })
+    .on('end', () => {
+      console.log('Extraction completed successfully.');
```
RaananW, line 33: this will happen async. as you don't define a callback to be executed when done, you can't 100% be sure when this has happened.
The process will stay working until reaching this point (or until and error occurs) but it might be good to catch them either using success/fail callbacks or wrapping them in a promise

Modules/@babylonjs/react-native-basekit/scripts/install.js
```diff
@@ -0,0 +1,174 @@
...
+  var site = getArgument('--brn-binary-site') ||
+              process.env.BRN_BINARY_SITE  ||
+              process.env.npm_config_brn_binary_site ||
+              (package.nodeBabylonConfig && package.nodeBabylonConfig.binarySite) ||
+              'https://github.com/CedricGuillemet/BabylonReactNative-1/releases/download';
+
+  return [site, package.version, binaryFilename].join('/');
+}
+
+const downloadExtractAndCache = (url, cachedFilePath) => {
```
RaananW, line 60: same as above - this function is async. Wrapping it in a promise will allow you to catch errors or when it is done at a higher level. Looking at the archiectuce, you are always ending with either cache or this function, so it should work as expected, but it should be a good practice to notify the parent caller when an async function is done.

RaananW (commented):
just a few comments on the install script :-)

Modules/@babylonjs/react-native-basekit/scripts/install.js
CedricGuillemet: Good catch!

Modules/@babylonjs/react-native-basekit/scripts/install.js
CedricGuillemet: no, it will be replaced with Babylon github

Modules/@babylonjs/react-native-basekit/scripts/install.js
CedricGuillemet: Fixed

Modules/@babylonjs/react-native-basekit/scripts/install.js
CedricGuillemet: yes

Modules/@babylonjs/react-native-basekit/scripts/install.js
CedricGuillemet: reactNativeVersion can be '0.71.X', would it coerce to 0.71 ?

Modules/@babylonjs/react-native-basekit/scripts/install.js
CedricGuillemet: Should be better now...I hope :)

Modules/@babylonjs/react-native-basekit/scripts/install.js
CedricGuillemet: Reworked as well.

Modules/@babylonjs/react-native-basekit/scripts/install.js
RaananW: hm.  That's incorrect comparison. It is true that `"0.7.X > "0.7"` is true, but this is false: `"0.10.X > "0.7"`. It works well until 9 because of the way string comparison works in JS.

Modules/@babylonjs/react-native-basekit/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+This quick overview will help you understand the constructs provided by Babylon React Native and how to use them in a React Native application.
+
+### Dependencies
+
+This package has several **peer dependencies**. If these dependencies are unmet, `npm install` will emit warnings. Be sure to add these dependencies to your project.
+
+### Installation Configuration
+
+Downloaded dependencies are cache by default in directory specifiec by `npm_config_cache`. It's possible to specify a new cache directory by setting `npm_config_babylon_binary_cache` environment variable.
```
ryantrem, line 13:
```suggestion
Downloaded dependencies are cached by default in directory specified by `npm_config_cache`. It's possible to specify a new cache directory by setting `npm_config_babylon_binary_cache` environment variable.
```

Modules/@babylonjs/react-native-basekit/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+Also, it's possible to set the source of these downloaded dependencies. By default, it's using main Github repo releases: `https://github.com/BabylonJS/BabylonReactNative/releases/download`.
+Other possibilities are:
+- By command argument for `npm install` : `--brn-binary-site=https://mysite.com/downloads`
```
ryantrem, line 17: It looks like the `--brn...` is supposed to be an argument to `npm install`, but I'm not sure why these are separated as if two different commands and what the colon is supposed to mean. What am I missing?

Modules/@babylonjs/react-native-basekit/README.md
```diff
@@ -0,0 +1,120 @@
...
+- By environment variable `export npm_config_brn_binary_site=https://mysite.com/downloads`
```
ryantrem, line 18: What does this mean? Is `export` some command? If I try to run this locally it just says export is not a recognized command.

Modules/@babylonjs/react-native-basekit/README.md
```diff
@@ -0,0 +1,120 @@
...
+- By configuration in package.json: `nodeBabylonConfig: { binarySite: 'https://mysite.com/downloads'}`
```
ryantrem, line 19: Where in the package.json? Like this is a top level json element or something?

Modules/@babylonjs/react-native-basekit/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+### iOS Configuration
+
+The minimum deployment target version is 12. This must be set as `iOS Deployment Target` in the consuming project's `project.pbxproj`, and must also be set as `platform` in the consuming project's `podfile`.
+
+### Platform Native Packages
+
+Babylon React Native supports react-native from 0.69 to 0.71+. See project Github main readme for supported versions.
+
+### `useEngine`
```
ryantrem, line 35: Is this duplicated documentation? Is the plan that once we are fully on this new packaging scheme, we delete the old stuff?

Modules/@babylonjs/react-native-basekit/scripts/install.js
```diff
@@ -0,0 +1,227 @@
...
+        });
+    } else {
+      resolve(false);
+    }
+  });
+}
+
+function getArgument(name) {
+  const flags = process.argv.slice(2),
+    index = flags.lastIndexOf(name);
```
ryantrem, line 49: This seems unusual, maybe make it more standard?
```suggestion
  const flags = process.argv.slice(2);
  const index = flags.lastIndexOf(name);
```

Modules/@babylonjs/react-native-basekit/scripts/install.js
```diff
@@ -0,0 +1,227 @@
...
+
+  if (index === -1 || index + 1 >= flags.length) {
+    return null;
+  }
+
+  return flags[index + 1];
```
ryantrem, line 55: It seems like this is getting the next argument after the named arg passed into this function?

Modules/@babylonjs/react-native-basekit/scripts/install.js
```diff
@@ -0,0 +1,227 @@
...
+}
+
+function getBinaryUrl(package, binaryFilename) {
+  var site =
```
ryantrem, line 59:
```suggestion
  const site =
```

Modules/@babylonjs/react-native-basekit/scripts/install.js
```diff
@@ -0,0 +1,227 @@
...
+  return [site, package.version, binaryFilename].join("/");
+}
+
+function downloadExtractAndCache(url, cachedFilePath, cb) {
+  const options = {
+    followRedirects: true, // Follow HTTP 3xx redirects
+  };
+
+  https
+    .get(url, options, (response) => {
```
ryantrem, line 75:
```suggestion
  https.get(url, options, (response) => {
```

Modules/@babylonjs/react-native-basekit/scripts/install.js
```diff
@@ -0,0 +1,227 @@
...
+        const untar = tar.extract({ cwd: process.cwd() }); // Extract to the current working directory
+
+        // create cache directory to store download file
+        const dir = path.dirname(cachedFilePath);
+        try {
+          if (!fs.existsSync(dir)) {
+            fs.mkdirSync(dir, { recursive: true });
+          }
+        } catch (err) {
+          reject(err);
```
ryantrem, line 90: Are all these `reject(err)` calls supposed to be `cb(err)`? It doesn't seem like there is a `reject` in this context since it is not `Promise` based. I wonder if it would be better to write this in TypeScript and have a step to compile it to JavaScript?

Modules/@babylonjs/react-native-basekit/scripts/install.js
```diff
@@ -0,0 +1,227 @@
...
+        console.error("Error install BabylonReactNative archive:", err);
+        process.exit(1);
+      });
+  } else {
+    console.error("No react-native version found for BabylonReactNative.");
+    process.exit(1);
+  }
+}
+
+install();
```
ryantrem, line 227: Isn't this a fire and forget async call? Will the npm install somehow wait for this operation to complete, or could this still be running after the `npm install` thinks it finishes? I thought I read somewhere it is possible for npm post install scripts to be async...

Modules/@babylonjs/react-native-basekit/README.md
CedricGuillemet: It's just me trying to separate the command and the parameter. it actually makes more sense to join both.

Modules/@babylonjs/react-native-basekit/README.md
CedricGuillemet: added this command is specific to mac/linux

Modules/@babylonjs/react-native-basekit/README.md
CedricGuillemet: yes, it's duplicated from @babylonjs/react-native readme. long term is indeed to replace old legacy packages by this one.

Modules/@babylonjs/react-native-basekit/scripts/install.js
CedricGuillemet: Fixed

Modules/@babylonjs/react-native-basekit/scripts/install.js
CedricGuillemet: Fixed

Modules/@babylonjs/react-native-basekit/scripts/install.js
CedricGuillemet: That's Prettier way to format this line. It's weird indeed.

Modules/@babylonjs/react-native-basekit/scripts/install.js
CedricGuillemet: In practice, it looks like npm waits for the install to finish. Can you confirm @RaananW ?

Modules/@babylonjs/react-native-basekit/README.md
CedricGuillemet: I've taken the code from node-sass but I can't find documentation describing in details how it works. I'm removing it for now.

Modules/@babylonjs/react-native-basekit/scripts/install.js
CedricGuillemet: Yes. Fixed.

Modules/@babylonjs/react-native-basekit/scripts/install.js
CedricGuillemet: I think it's be better to use TS. For a future PR? I'm pretty sure this new package will needed adjustments and bug fixes.

Modules/@babylonjs/react-native-basekit/README.md
CedricGuillemet: fixed

Modules/@babylonjs/react-native-basekit/scripts/install.js
```diff
@@ -0,0 +1,225 @@
...
+  const site =
+    getArgument("--brn-binary-site") ||
+    process.env.BRN_BINARY_SITE ||
+    process.env.npm_config_brn_binary_site ||
+    "https://github.com/BabylonJS/BabylonReactNative/releases/download";
+
+  return [site, package.version, binaryFilename].join("/");
+}
+
+function downloadExtractAndCache(url, cachedFilePath, cb) {
```
bghgary, line 68: It seems like this file could benefit from using async/await instead of callbacks. It would remove a bunch of the callback chaining and make error handling better.

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/BabylonReactNative.vcxproj
```diff
@@ -22,12 +22,15 @@
   <PropertyGroup Label="ReactNativeWindowsProps">
     <!-- First try and resolve react-native-windows package at a relative location when installing the @babylonjs/react-native package in your project -->
     <ReactNativeWindowsDir Condition="'$(ReactNativeWindowsDir)' == '' And Exists('$(ProjectDir)..\..\..\..\..\node_modules\react-native-windows\package.json')">$(ProjectDir)\..\..\..\..\..\node_modules\react-native-windows\</ReactNativeWindowsDir>
+	<!-- Then try and resolve react-native-basekit  package at a relative location when installing the @babylonjs/react-native-basekit package in your project-->
+	<ReactNativeWindowsDir Condition="'$(ReactNativeWindowsDir)' == '' And Exists('$(ProjectDir)..\..\..\..\..\node_modules\react-native-basekit\package.json')">$(ProjectDir)\..\..\..\..\..\node_modules\react-native-basekit\</ReactNativeWindowsDir>
```
bghgary, line 26: nit: Tab vs spaces issues in this file

.github/workflows/append_release.yml
```diff
@@ -0,0 +1,189 @@
+name: 'Append Release'
```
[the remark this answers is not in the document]
CedricGuillemet: In the case of a release, assets are appended thanks to the tag name. In case of a preview, a release is created on the fly.

.github/workflows/append_release.yml
```diff
@@ -0,0 +1,189 @@
...
+          name: 'Assembled'
+          path: Package/react-native
+      - name: Compress Package react-native
+        uses: a7ul/tar-action@v1.1.3
+        id: compressreact-native
+        with:
+          command: c
+          cwd: Package/react-native/
+          files: | # all files except package.json and readme.md
+            BabylonModule.d.ts
```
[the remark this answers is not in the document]
CedricGuillemet: All this is to avoid copying (and overwriting) package.json. I agree it's error-prone. atm, I only see more convoluted ways (a prepass that copies files, a js script that lists files,...)

.github/workflows/append_release.yml
```diff
@@ -0,0 +1,189 @@
...
+            react-native.tar.gz
+            Windows0.69.tar.gz
+            Windows0.70.tar.gz
+            Windows0.71.tar.gz
+            iOSAndroid0.69.tar.gz
+            iOSAndroid0.70.tar.gz
+            iOSAndroid0.71.tar.gz
+          tag_name: ${{ inputs.release-name }}
+          name: ${{ inputs.release-name }}
+          body: "Preview release for ${{ inputs.release-name }}"
```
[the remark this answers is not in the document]
CedricGuillemet: for non-preview, this body is filled by the maintainer. It's the release description with changelog etc..

Modules/@babylonjs/react-native-basekit/scripts/install.js
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  return flags[index + 1];
+}
+
+function getBinaryUrl(package, binaryFilename) {
+  const site =
+    getArgument("--brn-binary-site") ||
+    process.env.BRN_BINARY_SITE ||
+    process.env.npm_config_brn_binary_site ||
```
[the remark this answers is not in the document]
CedricGuillemet: I've seen that in nbode-sass and believed it was standard practice.

Modules/@babylonjs/react-native-basekit/scripts/install.js
CedricGuillemet: I guess so. Usage of type script might help as well.

Modules/@babylonjs/react-native-windows/windows/BabylonReactNative/BabylonReactNative.vcxproj
CedricGuillemet: fixed.

Modules/@babylonjs/react-native-basekit/scripts/install.js
bghgary: Meh, maybe. TypeScript is probably a bit overkill. Anyways, it's not a big deal either way.

.github/workflows/append_release.yml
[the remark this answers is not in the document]
bghgary: If there were a way to exclude files for tar, would that work?

Modules/@babylonjs/react-native-basekit/scripts/install.js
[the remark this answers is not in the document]
bghgary: I looked at node-sass and don't see what you mean. Where do you see duplicate environment variables?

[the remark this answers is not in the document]
CedricGuillemet: https://github.com/sass/node-sass/blob/87f38998da059d13a87a7fc9f3d29d5728128e2c/lib/extensions.js#L249

.github/workflows/append_release.yml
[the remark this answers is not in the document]
CedricGuillemet: yes, we do a PR to https://github.com/a7ul/tar-action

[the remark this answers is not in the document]
bghgary: Would it be okay to call tar from the shell?

CedricGuillemet:
Replaced with https://github.com/BabylonJS/BabylonReactNative/pull/665

PR #616, opened 2023-10-23
fix podspec include path for BabylonInterop

CedricGuillemet:
add include path in podspec so including BabylonNative works with 'Legacy' BRN package or with https://github.com/BabylonJS/BabylonReactNative/pull/615


bghgary (approved):
I don't know much about podspec files, so maybe @ryantrem can take a quick look before we merge.

ryantrem (approved):

Issue #343, opened 2022-02-17
Main thread is blocked when rendering


bghgary:
See https://github.com/BabylonJS/BabylonReactNative/blob/b3716d9dd7538ecf56ae670a311fba5b0d4b90b8/Modules/%40babylonjs/react-native/shared/BabylonNative.cpp#L116-L119

The code is currently calling `FinishRenderingCurrentFrame` immediately following `StartRenderingCurrentFrame`. `FinishedRenderingCurrentFrame` will immediately block waiting for the JS thread to finish doing its part. Ideally, these calls should be backwards. `StartRenderingCurrentFrame` should be called as soon as possible and `FinishRenderingCurrentFrame` should be called as late as possible. There should be a small gap between `FinishRenderingCurrentFrame` and the next `StartRenderingCurrentFrame`.

bghgary:
Closing in favor of [#1416](https://github.com/BabylonJS/BabylonNative/issues/1416)

PR #630, opened 2024-01-04
Swap start/finish order for rendering

ryantrem:
In this change, the order of start/finish for rendering (`Graphics::Device` and `Graphics::DeviceUpdate`) is swapped (similar to what was already done in the BN Playground app). This is more efficient as it results in less blocked time on the main thread, and should also allow the JS thread to start its per-frame work earlier. Without these changes, there can be more "fighting" for CPU time between BN and platform UI. I tested these changes by adding a large React Native `FlatList` into the main screen of the Playground app, and testing how scrolling the `FlatList` while interacting with the scene affected performance. Just looking at the frame rate, it seems like I get about a 10% improvement in performance (both Android and iOS). I expect we'd get better gains on Android with future work moving the rendering off the main thread entirely.

**Testing**
I tested pretty thoroughly with the Playground app on Android and iOS. I wasn't able to get a Windows build working (haven't tried in a long time), so I haven't tested on Windows (could use some help here 🙏🏻).

### DIFF
--- a/Modules/@babylonjs/react-native/shared/BabylonNative.cpp
+++ b/Modules/@babylonjs/react-native/shared/BabylonNative.cpp
@@ -103,8 +103,11 @@ namespace BabylonNative
             }
             else
             {
-                g_update->Finish();
-                g_graphicsDevice->FinishRenderingCurrentFrame();
+                if (m_isRenderingEnabled)
+                {
+                    g_update->Finish();
+                    g_graphicsDevice->FinishRenderingCurrentFrame();
+                }
 
                 g_graphicsDevice->UpdateWindow(m_graphicsConfig.Window);
                 g_graphicsDevice->UpdateSize(m_graphicsConfig.Width, m_graphicsConfig.Height);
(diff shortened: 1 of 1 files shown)


CedricGuillemet:
I'm still able to test Windows build. I'll check it this week.

CedricGuillemet:
I guess it fixes https://github.com/BabylonJS/BabylonReactNative/issues/343

matthargett:
Can this be rebased and retested?

ryantrem:
@CedricGuillemet you did test this a while back on Windows and something wasn't working, correct? Do you recall what the issue was? Can you add some notes to this PR discussion?

CedricGuillemet:
> @CedricGuillemet you did test this a while back on Windows and something wasn't working, correct? Do you recall what the issue was? Can you add some notes to this PR discussion?

Sorry, I don't remember testing this PR. I can add that to my todo list

ryantrem:
Found a private chat between us back in January and you said:
> I tried to build&run your fix-render-order branch with PG 0.71 on Windows. It builds but crashes at startup. no log, no info on why it happens

No additional discussion after that, so I think that is as far as we got with testing on Windows and then the PR was put on hold.

PR #633, opened 2024-01-23
Notice html distribution

CedricGuillemet:
Add `NOTICE.html` from repo root folder inside every npm package next to its readme.
I've run a dry-run publish, checked every .tgz and they all contains the notice file. https://github.com/BabylonJS/BabylonReactNative/actions/runs/7628535866


bghgary (commented):
I will update this file with the proper contents in a separate change.

bghgary (approved):

PR #636, opened 2024-01-29
Remove RN 064,065 from builds/packages

CedricGuillemet:
Playground/PackageTest folders are kept for a few more release (just in case).
PackageTest should completely be removed once RNTA is up and running


ryantrem:
Will it be a later change to remove the 0.64 and 0.65 Playground directories/apps?

Also, when you say "PackageTest should completely be removed once RNTA is up and running" is the idea that we will be able to use RNTA to also swap between the local BRN package and a published BRN package? @kelset

kelset:
I think we'll have more clarity on that once I figure out how to make RNTA work in the repo - the local build issues we chatted about offline delayed me a bit, but I'm trying to make progress this week

CedricGuillemet:
switching to RNTA instead. Closing

PR #639, opened 2024-02-13
Markdown to HTML in gulp script

CedricGuillemet:
gulp script that converts a .MD file to Notice.html that can then be packaged with the npm.
As a test, this script converts the readme file.

Result:

image

TODO:
- get the notice.md file
- add the buildNotice step into the build/package process

### DIFF
--- a/Package/gulpfile.js
+++ b/Package/gulpfile.js
@@ -8,12 +8,35 @@ const shelljs = require('shelljs');
 const rename = require('gulp-rename');
 const glob = require('glob');
 const chalk = require('chalk');
+const marked = require('marked');
 
 let assemblediOSAndroidDir = 'Assembled-iOSAndroid';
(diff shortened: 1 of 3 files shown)


Package/gulpfile.js
```diff
@@ -8,13 +8,36 @@ const shelljs = require('shelljs');
...
+const marked = require('marked');
 
 let assemblediOSAndroidDir = 'Assembled-iOSAndroid';
 let assembledWindowsDir = 'Assembled-Windows';
 let basekitBuild = false;
 let cmakeBasekitBuildDefinition = '';
 let basekitPackagePath = '';
 
+
+const makeHTMLNotice = async () => {
```
ryantrem, line 20: This function is marked async but then uses fs with callbacks. I believe these days you can do `const fs = require('fs/promises')` instead and use Promise returning async functions.

PR #645, opened 2024-05-20
Update BN

### LINKED ISSUE #644: Dependency Conflict When Initializing App
**Describe the bug**
I was following this instruction to set up an app: https://github.com/BabylonJS/BabylonReactNativeSample/blob/main/CREATE.md
And it tells me that there is conflicting dependencies on @babylonjs/core.

I think the latest of @babylonjs/core is 7.3.1 but we are still using 6.14.0 here. Overring the @babylonjs/core to use 6.14.0 in package.json helps resolve, but I think the correct fix would be to upgrade this package to use 7.3.1.

**To Reproduce**
1. `npx react-native@latest init AwesomeProject`
2. `npm i @babylonjs/react-native @babylonjs/react-native-iosandroid-0-71 @babylonjs/core @babylonjs/loaders`
3. See the error
Screenshot 2024-04-19 at 21 57 07

**Expected behavior**
The npm install should complete without errors.

CedricGuillemet:
fixes #644 

Tested OK on Android except rendering (see  https://github.com/BabylonJS/BabylonNative/pull/1382 )
- Replaced Node 12 with 18 for publishing (why 12 before?).
- Xcode 15 instead of 13
- Fixed library path in vcxproj
- up BN to https://github.com/BabylonJS/BabylonNative/commit/59798a738d5c785d9e3eff0863854034e075f29f and bjs version must be >= 5.53.1


Modules/@babylonjs/react-native/package.json
```diff
@@ -28,15 +28,15 @@
     "semver": "^7.3.2"
   },
   "peerDependencies": {
-    "@babylonjs/core": "^6.14.0",
+    "@babylonjs/core": "^7.6.2",
```
ryantrem, line 31: It doesn't seem like we want this change... this is saying BRN requires at least BJS 7.6.2, but we document the min as 5.53.1. Seems like this should actually be ^5.53.1, right?

CedricGuillemet: yes! I'll fix that!

ryantrem (approved):
Did you run through the manual test script for both iOS and Android?

CedricGuillemet:
> Did you run through the manual test script for both iOS and Android?

not yet. I need to do a dry run of npm publish as well.

Modules/@babylonjs/react-native/package.json
CedricGuillemet: actually, because of non-float support in vertex buffer. bjs version will be >=7.7.1

ryantrem: Really? So this is basically a breaking change, where if anyone wants to update BRN, they are going to have to update BJS all the way to 7.7.1 as well? @bghgary

CedricGuillemet: I've update BN to the latest and works fine with BJS 5.53.1

ryantrem (approved):

PR #647, opened 2024-06-17
Use patch to select RN version for RNTA

CedricGuillemet:
WIP

use a git patch to select RN and apply dependency changes.


.github/workflows/pr.yml
```diff
@@ -60,267 +24,29 @@ jobs:
...
-      - name: Checkout Repo
-        uses: actions/checkout@v2.3.3
-        with:
-          submodules: 'recursive'
-      - name: Setup Watchman
-        run: brew install watchman
-  
-      - name: NPM Install (Playground)
+      - name: Patch BRNPlayground for RN ${{ matrix.react-native-version }}
+        run: git apply rnta-${{ matrix.react-native-version }}.patch --allow-empty
```
ryantrem, line 28: Can we do this as part of a script that is easy to run locally? Like the `npm run select 0.71` we currently have?

CedricGuillemet:
Replaced with https://github.com/BabylonJS/BabylonReactNative/pull/665

PR #662, opened 2024-11-20
fix: RNPermissions error in Playground

okwasniewski:
This PR fixes `RNPermissions` error in Playground. Currently, when launching the BRNPlayground from master the below error is thrown:

```
Invariant Violation: TurboModuleRegistry.getEnforcing(...): 'RNPermissions' could not be found. Verify that a module by this name is registered in the native binary.Bridgeless mode: false. TurboModule interop: false. Modules loaded: {"NativeModules":["UIManager"],"TurboModules":["PlatformConstants","LogBox","SourceCode","Timing","AppState","BlobModule","WebSocketModule","DevSettings","DevToolsSettingsManager","Networking","Appearance","DevLoadingView","HeadlessJsTaskSupport","DeviceInfo","DeviceEventManager"],"NotFound":["NativePerformanceCxx","NativePerformanceObserverCxx","RedBox","BugReporting","RNPermissions"]}, js engine: hermes
```

This PR addresses the issue.

**Screenshots**

CleanShot 2024-11-20 at 12 47 52@2x

**Documentation**

N/A

**Testing**

Launch test app from master


Modules/@babylonjs/react-native/package.json
```diff
@@ -43,7 +43,7 @@
     "@types/jest": "^25.2.1",
     "@types/react": "^16.9.32",
     "@types/react-native": "^0.63.1",
-    "@types/react-native-permissions": "^2.0.0",
+    "react-native-permissions": "^4.1.5",
```
ryantrem, line 46: Seems like we shouldn't need this if the types are already included in the package. Or were they only included as of 4.0?

Modules/@babylonjs/react-native/package-lock.json
ryantrem, line 1: Why is so much changed in package-lock? Did you upgrade a bunch of other packages as well? If so I think that is fine if everything is working, just trying to understand.

Modules/@babylonjs/react-native/package.json
okwasniewski: So for some reason when running in new architecture, the `@babylonjs/react-native` package needs the implementation of `react-native-permissions` in it's corresponding `node_modules` folder without it the mentioned error gets thrown. Also types are included as of 3+

Modules/@babylonjs/react-native/package-lock.json
okwasniewski: @ryantrem I updated other dependencies now, looks like react-native-permissions had peer dependency issue with @types/react 16.9 which should be updated... :)

I upgraded this for all 3 packages, just to make sure you can checkout my branch and try installing dependencies too but there shouldn't be any peer deps issues now.

Modules/@babylonjs/react-native/package.json
okwasniewski: It looks like it's a standard practice to add your peer dependencies as dev dependencies in development. 

Here is an example of react native builder bob (popular create-react-native-library) which adds `react-native` as dev dependency: https://github.com/callstack/react-native-builder-bob/blob/9a025b2a89c7790b74862bf95ccab4b6a388e2cb/packages/create-react-native-library/templates/common/%24package.json#L90

ryantrem (approved):

PR #663, opened 2024-11-26
fix: add support for merged so libraries

okwasniewski:
**Describe the change**

This PR adds support for React Native 0.76 merged .so libraries, reference: https://github.com/react-native-community/discussions-and-proposals/discussions/816

**Documentation**

N/A

**Testing**

Try to build the app for 0.76

The app is built properly but Im getting corrupted stack issues, both on 0.75 (without merged so) and 0.76 (with merged so) but this is mostly build-related change so verifying if the library builds properly is the most important step. I'm looking into the corrupted stack issues


okwasniewski:
This PR will become irrelevant once AAR support is included so probably can be closed

PR #736, opened 2026-07-30
Update BN commitid

CedricGuillemet:
## Summary

- Update the packaged BabylonNative source to `17d3e4e6b1233547d7ccfdf89da04c327340f8fc`.
- Adapt package assembly and native integration to current BabylonNative dependencies and build behavior.
- Keep Babylon.js compatibility at 9.0.0.
- Remove the unsupported Android `x86` ABI. Current bgfx/bx sources require SSE4.x intrinsics, which exceed Android's 32-bit x86 ABI baseline; consumer-provided `x86` entries are skipped with a warning.

## Validation

- Android build and package-consumer build pass in CI.
- iOS build and package-consumer build pass in CI.
- TypeScript and package jobs pass in CI.
- Installed and smoke-tested the Playground on a connected Pixel 9a (arm64-v8a).

## Follow-up

Windows project updates are included for the BabylonNative dependency changes, but the Windows build has not been tested because this repository's PR CI does not currently include Windows. Windows verification will be handled separately.


Modules/@babylonjs/react-native/windows/BabylonReactNative/BabylonReactNative.vcxproj
```diff
@@ -207,8 +206,6 @@
         SPIRVd.lib;
         UrlLib.lib;
         Window.lib;
```
bghgary, line 209: Removing these is right - BabylonNative dropped libwebp for bimg's bundled simplewebp decoder, so `libwebp`/`sharpyuv` no longer exist as targets. But BRN CI has no Windows job (android, iOS, typescript, package, test-package-*), so nothing here exercises the Windows link. Has it been built locally?

Modules/@babylonjs/react-native/android/build.gradle
```diff
@@ -39,7 +39,16 @@ apply plugin: 'kotlin-android'
 
 def reactNativeArchitectures() {
   def value = project.getProperties().get("reactNativeArchitectures")
-  return value ? value.split(",") : ["armeabi-v7a", "x86", "x86_64", "arm64-v8a"]
+  def architectures = value ? value.split(",").collect { it.trim() } : ["armeabi-v7a", "x86_64", "arm64-v8a"]
+  // The 32-bit x86 ABI is not supported: bgfx's bx uses SSE4.x intrinsics
+  // (_mm_blendv_ps, _mm_round_ps), which are above the Android x86 ABI baseline
+  // of SSSE3 and fail to compile. Drop it rather than break the whole build for
+  // consumers still carrying x86 in the React Native default property.
+  def unsupported = architectures.findAll { it == "x86" }
```
bghgary, line 47: Dropping `x86` here changes the published module's behaviour, not just the Playground's: a consumer with the React Native default `reactNativeArchitectures` silently loses an ABI (warning only). Correct given bx needs SSE4.x, but worth saying so in the PR description - the title reads as a commit-id bump, and the body is empty.

Modules/@babylonjs/react-native/README.md
```diff
@@ -74,7 +74,7 @@ Babylon.js minimal version:
 
 |BabylonReactNative version | Babylon.js version | BabylonNative commit |
 | ----------- | ------------------------ | --- |
-|2.0.2 | 9.0.0 | ce2edf0851e0c8483559832a4f8eb9d39c6b2f53
+|2.0.2 | 9.22.2 | 17d3e4e6b1233547d7ccfdf89da04c327340f8fc
```
bghgary: Should the `@babylonjs/core` peer dependency also be raised to `>=9.22.2`? This table now declares 9.22.2 as the minimum, but `package.json` and `VersionValidation.ts` still accept `>=9.0.0`, so older versions pass installation and runtime validation.

bghgary (approved):
[Reviewed by Copilot on behalf of @bghgary]

LGTM with comments inline.

Modules/@babylonjs/react-native/android/build.gradle
CedricGuillemet: Documented the published-module behavior change in the PR description, including why `x86` is skipped and that consumers receive a warning.

Modules/@babylonjs/react-native/windows/BabylonReactNative/BabylonReactNative.vcxproj
CedricGuillemet: Not yet. Windows testing is outside this PR because there is no Windows PR CI job; I documented that limitation and will revisit Windows verification separately.

Modules/@babylonjs/react-native/README.md
CedricGuillemet: Reverted the Babylon.js update and restored the Playground and compatibility table to 9.0.0, so the existing peer dependency and runtime validation remain consistent.

PR #754, opened 2026-09-09
Fix Windows embedded source build

CedricGuillemet:
## Summary

- package Babylon Native's embedded source tree and build it from Android, iOS, and Windows consumer projects
- make the Windows MSBuild bridge run CMake only after React Native Windows generates its C++/WinRT projections
- pass RNW's generated projection tree to every CMake target, inherit RNW's `Microsoft.Windows.CppWinRT` package policy, and remove the duplicate legacy `packages.config` integration
- compile the embedded Windows graph as C++20 and isolate CMake outputs by configuration, toolset, generator, and platform while keeping paths below legacy MSBuild limits
- add blocking Windows source-tree and packaged-tarball consumer builds
- keep the Playground controls visible and interactive above native rendering surfaces on Android and Windows, report XR startup errors, and contribute optional ARCore manifest metadata

## Windows C++/WinRT compatibility

The generated React Native Windows projection tree is the single source of C++/WinRT headers for the embedded CMake build. The build fails if that tree is unavailable instead of silently falling back to a potentially different Windows SDK projection version. This keeps BRN aligned with the RNW version selected by the consuming application rather than pinning BRN to one machine-specific SDK projection.

Windows consumer CI keeps C++20 scoped to BRN's embedded CMake graph and the generated test application; it does not change RNW source projects. Autolinked native modules are restored and compiled independently against the consumer-selected RNW NuGet packages, then the generated x64 application is compiled through C++/WinRT and XAML and linked as `ReactTestApp.exe`. AppX packaging is intentionally excluded because it requires resource artifacts from recursively built RNW source projects, which are outside the NuGet consumer graph being validated.

## Validation

- all jobs passed in PR Build run `34877434422`, including the Windows source-tree consumer and packaged-tarball consumer
- both Windows jobs compile the embedded Babylon Native sources, autolinked native modules, generated C++/WinRT/XAML application code, and link the x64 application
- built and installed the Windows Playground package locally and visually confirmed all controls render correctly
- built the full Android Debug app with `libReactNativeBabylon.so`, installed it on a physical Pixel 9a, and confirmed the native rendering surface no longer overlaps the controls
- confirmed Android XR startup reaches the camera permission flow without the previous missing `com.google.ar.core` manifest failure
- Playground regression suite passes: controls are layered above `EngineView`, and rejected XR startup and shutdown are surfaced to the user
- merged the latest `origin/master` into the PR branch


Modules/@babylonjs/react-native/windows/BabylonReactNative/BabylonReactNative.vcxproj
```diff
@@ -99,7 +106,7 @@
     <Import Project="PropertySheet.props" />
   </ImportGroup>
   <ImportGroup Label="ReactNativeWindowsPropertySheets">
-    <Import Project="$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.Uwp.CppLib.props" Condition="Exists('$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.Uwp.CppLib.props')" />
+    <Import Project="$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.CppLib.props" Condition="Exists('$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.CppLib.props')" />
```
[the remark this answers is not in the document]
CedricGuillemet: Added conditional fallbacks for both `Microsoft.ReactNative.CppLib.*` and `Microsoft.ReactNative.Uwp.CppLib.*`, including validation that fails only when neither RNW layout is available. This preserves the documented RNW 0.81.4+ compatibility.

.github/workflows/test_windows.yml
```diff
@@ -0,0 +1,60 @@
...
+        working-directory: ./Apps/Playground
+
+      - name: Build packaged sources
+        shell: pwsh
+        run: |
+          $project = Resolve-Path ..\node_modules\@babylonjs\react-native\windows\BabylonReactNative\BabylonReactNative.vcxproj
+          $solutionPath = (Resolve-Path Playground.sln).Path
+          $solutionDir = Split-Path $solutionPath
+          $solutionFileName = Split-Path $solutionPath -Leaf
+          MSBuild $project /restore /m /p:Platform=x64 /p:Configuration=Release /p:SolutionPath=$solutionPath /p:SolutionDir=$solutionDir /p:SolutionFileName=$solutionFileName /p:WindowsTargetPlatformVersion=${{ inputs.WindowsTargetPlatformVersion }} /p:UseExperimentalNuget=true /p:WindowsAppSDKVerifyTransitiveDependencies=false /v:minimal
```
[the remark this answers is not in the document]
CedricGuillemet: The packaged workflow now builds the generated `Playground.sln` after the direct library build, so the installed tarball is exercised through the consumer app's integration, linking, and packaging path. It also runs `npm ci` before overlaying the tarball so the checked-in lockfile is honored.

Apps/Playground/App.tsx
```diff
@@ -136,44 +138,52 @@ const EngineScreen: FunctionComponent<ViewProps> = (props: ViewProps) => {
...
+      // TODO: Figure out why getFrontPosition stopped working
+      //box.position = (scene.activeCamera as TargetCamera).getFrontPosition(2);
+      const cameraRay = scene.activeCamera!.getForwardRay(1);
+      rootNode.position = cameraRay.origin.add(
+        cameraRay.direction.scale(cameraRay.length)
+      );
+      rootNode.rotate(Vector3.Up(), 3.14159);
+    } catch (error) {
+      const message = error instanceof Error ? error.message : String(error);
+      setXrError(`Unable to start XR: ${message}`);
```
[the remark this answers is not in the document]
CedricGuillemet: The handler now captures whether the invocation is starting or stopping XR before entering the try block and reports the matching action. Added regression coverage for a rejected `exitXRAsync()`.

.github/workflows/windows.yml
```diff
@@ -11,39 +11,36 @@ on:
...
-      - name: Windows build Playground
-        run: MSBuild /p:Platform="x64" /p:Configuration="Release" /m Playground.sln /v:detailed /p:WindowsTargetPlatformVersion=${{ inputs.WindowsTargetPlatformVersion }}
+      - name: Build Babylon React Native
+        shell: pwsh
+        run: |
+          $project = Resolve-Path ..\node_modules\@babylonjs\react-native\windows\BabylonReactNative\BabylonReactNative.vcxproj
+          $solutionPath = (Resolve-Path Playground.sln).Path
+          $solutionDir = Split-Path $solutionPath
+          $solutionFileName = Split-Path $solutionPath -Leaf
+          MSBuild $project /restore /m /p:Platform=x64 /p:Configuration=Release /p:SolutionPath=$solutionPath /p:SolutionDir=$solutionDir /p:SolutionFileName=$solutionFileName /p:WindowsTargetPlatformVersion=${{ inputs.WindowsTargetPlatformVersion }} /p:UseExperimentalNuget=true /p:WindowsAppSDKVerifyTransitiveDependencies=false /v:minimal
```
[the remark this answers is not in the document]
CedricGuillemet: Added a separate blocking `Playground.sln` build after the direct BRN project build. This retains focused native bridge coverage while also compiling and packaging the generated RNW consumer.

RaananW (dismissed):
> 🤖 This review comment was created by an AI agent.

Thanks for adding direct Windows consumer checks. Both Windows jobs fail on the current head. The solution-wide `LanguageStandard=stdcpp20` setting at [.github/workflows/windows.yml:48-52](https://github.com/BabylonJS/BabylonReactNative/blob/524725ae02ad864f72424ec4694d09edc2da28f7/.github/workflows/windows.yml#L48-L52) and [.github/workflows/test_windows.yml:63-67](https://github.com/BabylonJS/BabylonReactNative/blob/524725ae02ad864f72424ec4694d09edc2da28f7/.github/workflows/test_windows.yml#L63-L67) also changes `Microsoft.ReactNative.vcxproj`. RNW then mixes `std::coroutine_handle` and `std::experimental::coroutine_handle`, so both consumer builds fail. Please keep C++20 scoped to the projects that need it and restore green Windows checks.

.github/workflows/windows.yml
```diff
@@ -11,39 +11,43 @@ on:
...
         working-directory: ./Apps/Playground/windows
 
-        
\ No newline at end of file
+      - name: Build Windows consumer
+        shell: pwsh
+        run: |
+          $solutionPath = (Resolve-Path Playground.sln).Path
+          MSBuild $solutionPath /restore /m /p:Platform=x64 /p:Configuration=Release /p:LanguageStandard=stdcpp20 /p:WindowsTargetPlatformVersion=${{ inputs.WindowsTargetPlatformVersion }} /p:UseExperimentalNuget=true /p:WindowsAppSDKVerifyTransitiveDependencies=false /v:minimal
```
bghgary: Both new Windows jobs fail at this solution-wide override: it applies C++20 to `Microsoft.ReactNative.vcxproj`, whose generated headers and Hermes sources then disagree between `std::coroutine_handle` and `std::experimental::coroutine_handle`. Keep C++20 scoped to BRN/CMake targets instead of passing `/p:LanguageStandard=stdcpp20` to either solution build, and restore both Windows checks.

Apps/Playground/__tests__/App.test.tsx
```diff
@@ -0,0 +1,155 @@
+import React from "react";
```
bghgary: No pull-request workflow invokes this new suite, so the XR error-handling and control-layering regressions can merge even when these tests fail. Run the Playground Jest suite in a blocking workflow.

Apps/Playground/generate-windows-test-app.mjs
```diff
@@ -0,0 +1,15 @@
+import path from "node:path";
+
+import { generateSolution } from "./node_modules/react-native-test-app/windows/app.mjs";
+
+const destination = path.resolve(process.argv[2] ?? "windows");
+const error = await generateSolution(destination, {
+    autolink: true,
+    useFabric: false,
```
bghgary: This PR adds architecture-aware React Native Windows project imports, but both Windows jobs force the legacy `useFabric: false` configuration even though `react-native-test-app` defaults React Native Windows 0.80+ to the new architecture. Add a Fabric Windows consumer build so the new Composition/Fabric production path is validated.

RaananW (dismissed):
🤖 This review comment was created by an AI agent.

Thank you for adding direct Windows consumer checks. The current head is unchanged since the earlier review, and both Windows jobs still fail. The solution-wide `LanguageStandard=stdcpp20` setting also changes `Microsoft.ReactNative.vcxproj`, which mixes standard and experimental coroutine types. Please keep C++20 scoped to the projects that need it and restore green Windows checks.

ryantrem (approved):

.github/workflows/windows.yml
```diff
@@ -8,42 +8,136 @@ on:
...
+            "/p:Configuration=Release",
+            "/p:SolutionPath=$solutionPath",
+            "/p:SolutionDir=$solutionDir",
+            "/p:SolutionFileName=$solutionFileName",
+            "/p:WindowsTargetPlatformVersion=${{ inputs.WindowsTargetPlatformVersion }}",
+            "/p:WindowsTargetPlatformMinVersion=${{ inputs.WindowsTargetPlatformMinVersion }}",
+            "/p:ForceImportAfterCppTargets=$reactNativeNuGetTargetsPath",
+            "/p:WindowsAppSDKVerifyTransitiveDependencies=false",
+            "/p:AppxBundle=Never",
+            "/p:AppxPackage=false",
```
bghgary: Both Windows jobs now pass `AppxPackage=false`; their successful logs emit APPX1503 that the manifest is ignored because no app package is produced. This hides the missing React Native Windows runtime payloads exposed by the prior AppX failures. Restore an AppX-packaging check and source the required payloads from the NuGet package.

Apps/Playground/package-windows.json
```diff
@@ -20,7 +20,7 @@
     "react": "19.1.4",
     "react-native": "^0.81.4",
     "react-native-permissions": "^5.4.1",
-    "react-native-windows": "^0.81.4"
+    "react-native-windows": "^0.81.35"
```
bghgary: This raises the only Windows CI consumer from React Native Windows `^0.81.4` to `^0.81.35` while the README documents 0.81.4+ support. Compatibility-sensitive import and NuGet metadata changes can now break the documented minimum unnoticed. Add a blocking 0.81.4 build or update the documented minimum.

